PR #19, opened 2019-11-07
add CNN and LSTM classification models

jxmorris12:


Opdoop:
@jxmorris12 I changed dropout to 0.3 but didn't see the warning, do you have any reference that `Fix error where we ask for non-zero dropout with only 1 layer.`

Opdoop:
Never mind. I saw the warning when using a different methine. 🤷‍♂️

PR #82, opened 2020-04-29
seq2sick

jxmorris12:
depends on #80 

action items:
- add en->ro, en->fr, summarization datasets
- implement true seq2sick attack  using gradient descent


scripts/run_attack_args_helper.py
```diff
@@ -34,21 +34,28 @@
...
+    #
+    # Translation models
+    #
+    't5-en2fr':                 'textattack.models.translation.t5.T5EnglishToFrench',
+    't5-en2de':                 'textattack.models.translation.t5.T5EnglishToGerman',
+    't5-en2ro':                 'textattack.models.translation.t5.T5EnglishToRomanian',
+    #
+    # Summarization models
+    #
+    't5-summ':                  'textattack.models.summarization.T5Summarization',
```
uvafan, line 52: So we have a summarization model but no summarization datasets yet right

textattack/datasets/translation/translation_datasets.py
```diff
@@ -0,0 +1,23 @@
...
+        0 - Negative
+        1 - Positive
+
+    Args:
+        src_lang (str): source language
+        offset (int): example to start
+    
+    """
+    DATA_PATH = 'datasets/translation/NewsTest2013EnglishToGerman'
+    def __init__(self, offset=0):
```
uvafan, line 22: Offset argument isn't used?

textattack/datasets/translation/translation_datasets.py
```diff
@@ -0,0 +1,23 @@
+import gluonnlp
```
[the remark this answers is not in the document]
uvafan: Also add it to requirements.txt if you are

textattack/attack_recipes/cheng_2018_seq2sick.py
```diff
@@ -0,0 +1,38 @@
+"""
+    Cheng, Minhao, et al. 
+    
+    Seq2Sick: Evaluating the Robustness of Sequence-to-Sequence Models with 
+        Adversarial Examples
+    
+    ArXiv, abs/1803.01128.
+    
+    
+    This is a greedy re-implementation of the seq2sick attack method. It does
```
uvafan, line 10: I think until we implement gradient descent we should name it seq2sick_black_box or something, since it's not really the same method while the other attack recipes are supposed to emulate as close as possible.

uvafan:
One more thing to consider: I feel like the output isn't as clearly interpreted in translation. At first I thought the first line was the source language and the next two lines were the different translations. This doesn't happen with classification since one is obviously a label and one is a string. Maybe we should make it more clear which is the source language and which is the target.

scripts/benchmark_models.py
```diff
@@ -7,9 +7,9 @@
 
 import textattack.models as models
 
-def _cb(s): return textattack.shared.utils.color(str(s), color='blue', method='stdout')
-def _cg(s): return textattack.shared.utils.color(str(s), color='green', method='stdout')
-def _cr(s): return textattack.shared.utils.color(str(s), color='red', method='stdout')
+def _cb(s): return textattack.shared.utils.color_text_by_method(str(s), color='blue', method='stdout')
+def _cg(s): return textattack.shared.utils.color_text_by_method(str(s), color='green', method='stdout')
```
jxmorris12, line 11: need to change to `color_text`

textattack/attack_methods/attack.py
```diff
@@ -123,33 +123,33 @@ def _get_examples_from_dataset(self, dataset, num_examples=None, shuffle=False,
...
         Args:
-            dataset: An iterable of (label, text) pairs
+            dataset: An iterable of (text, ground_truth_output) pairs
             num_examples (int): the number of examples to return
             shuffle (:obj:`bool`, optional): Whether to shuffle the data
             attack_n (bool): If `True`, returns `num_examples` non-skipped
                 examples. If `False`, returns `num_examples` total examples.
         
         Returns:
             results (List[Tuple[Int, TokenizedText, Boolean]]): a list of
```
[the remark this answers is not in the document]
jxmorris12: good catch

textattack/datasets/translation/translation_datasets.py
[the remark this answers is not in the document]
jxmorris12: Thanks, good catch. Originally I was going to (it has the dataset built-in) but I decided it was too hefty to require users to import - we want to try and keep the amount of large packages required down, I think. So I "made" the dataset (just exported it to a .pkl object) and put it on S3. So, anyways, we don't need gluonnlp anymore - removing

scripts/run_attack_args_helper.py
jxmorris12: yep (see PR description)

textattack/datasets/translation/translation_datasets.py
jxmorris12: good catch

textattack/attack_recipes/cheng_2018_seq2sick.py
jxmorris12: ok good idea

textattack/datasets/translation/translation_datasets.py
```diff
@@ -0,0 +1,23 @@
...
+
+class NewsTest2013EnglishToGerman(TextAttackDataset):
+    """ Loads samples from newstest2013 dataset from the publicly available
+            WMT2016 translation task. (This is from the 'news' portion of 
+            WMT2016. See http://www.statmt.org/wmt16/ for details.)
+            
+        Samples are loaded as (sample, translation) tuples of string pairs.
+            Relies on the GluonNLP library for dataset loading.
+    
+    Labels:
```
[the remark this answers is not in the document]
jxmorris12: fixed, thanks

jxmorris12:
thanks for the review. I put some Asana tasks up related to your general comments.

PR #83, opened 2020-05-01
add pypi files to gitignore; add better language_tool dependency

jxmorris12:


setup.py
```diff
@@ -5,14 +5,19 @@
 
 setuptools.setup(
     name="textattack",
-    version="0.0.1",
+    version="0.0.1.7",
     author="QData Lab at the University of Virginia",
     author_email="jm8wx@virginia.edu",
     description="A library for generating text adversarial examples",
+    include_package_data=False,
```
uvafan, line 12: just curious, what do this and the changes below do haha

.gitignore
```diff
@@ -26,4 +26,8 @@ docs/_build/
 tensorflow-hub
 
 # for Macs
-.DS_Store
\ No newline at end of file
+.DS_Store
+
+# build outputs for PyPI
+build/
```
uvafan, line 32: don't we need to also remove these from the repo in addition to adding them to gitignore?

jxmorris12: thanks, I don't know how / why the `build/` files were in the repo before, but I removed the folder!

setup.py
jxmorris12: This lets you specify specifically which non-.py files should be included along with the code. If it wasn't specified, pip would put the config.json in some other folder, potentially far away from the code. We need the config file to be imported from the code via relative imports, so I changed some settings to allow that

PR #87, opened 2020-05-05
Logging updates; add Weights & Biases

jxmorris12:
- rename `AttackLogger` to `AttackLogManager`
- some other minor renames
- add Weights & Biases logging

### DIFF
--- a/textattack/shared/scripts/run_attack_args_helper.py
+++ b/textattack/shared/scripts/run_attack_args_helper.py
@@ -133,25 +133,25 @@ def get_args():
         default=[], choices=CONSTRAINT_CLASS_NAMES.keys(),
         help=('Constraints to add to the attack. Usage: "--constraints {constraint}:{arg_1}={value_1},{arg_3}={value_3}"'))
     
-    parser.add_argument('--out_dir', type=str, required=False, default=None,
+    parser.add_argument('--out-dir', type=str, required=False, default=None,
         help='A directory to output results to.')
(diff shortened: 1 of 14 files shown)


textattack/shared/utils.py
```diff
@@ -171,3 +171,46 @@ class ANSI_ESCAPE_CODES:
...
+    """
+    style_str = ''
+    for key in style_dict:
+        style_str += key + ': ' + style_dict[key] + ';'
+    return 'style="{}"'.format(style_str)
+    
+def html_table_from_rows(rows, title=None, header=None, style_dict=None):
+    # Stylize the container div.
+    if style_dict:
+        table_html = '<div {}>'.format(style_from_dict(style_dict))
```
uvafan, line 189: change to html_style_from_dict

textattack/shared/scripts/run_attack_args_helper.py
```diff
@@ -139,6 +139,9 @@ def get_args():
     parser.add_argument('--enable_visdom', action='store_true',
         help='Enable logging to visdom.')
     
+    parser.add_argument('--enable_wandb', action='store_true',
```
uvafan, line 142: we should change all these '_'s to '-'s, pretty sure I made the first one my b

textattack/shared/utils.py
jxmorris12: nice catch

PR #90, opened 2020-05-08
minor fixes: config.json, tests, travis

jxmorris12:
- model batch size, model cache size, and constraint cache size in config.json
- improvements to `local_tests`: print stderr, add a test, make another test shorter
- fix travis


local_tests/command_line_tests.py
```diff
@@ -50,10 +50,26 @@ def register_test(command, name=None, output_file=None, desc=None):
...
+#
+# test: run_attack non-overlapping output of class 2 on T5 en->de translation with
+#   attack_n set, using the WordSwapRandomCharacterSubstitution transformation 
+#   and greedy word swap, using edit distance constraint, on 6 samples
+#                   (takes about 100s)
+#
+register_test(('python -m textattack --attack-n --goal-function non-overlapping-output '
+    '--model t5-en2de --num-examples 6 --transformation word-swap-random-char-substitution '
+    '--constraints edit-distance:12 words-perturbed:max_percent=0.75 --attack greedy-word'), 
+    name='run_attack_nonoverlapping_t5en2de_randomcharsub_editdistance_wordsperturbed_greedyword',
```
uvafan, line 71: should we give these tests shorter names if we're trying to enable people to choose which ones to run from command line? (I'm not sure what do you think?)

local_tests/sample_outputs/run_attack_nonoverlapping_t5ende_editdistance_bleu.txt
```diff
@@ -0,0 +1,64 @@
...
+One [94mHhing[0m is certain: these new provisions will have a negative impact on voter turn-out.
+
+
+--------------------------------------------- Result 6 ---------------------------------------------
+In diesem Sinne werden die Maßnahmen das demokratische System der USA teilweise untergraben-->[91m[FAILED][0m
+In this sense, the measures will partially undermine the American democratic system.
+
+
+
+(0lqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqwqqqqqqqqk(B
```
uvafan, line 51: Isn't this an file logging bug? I checked and it isn't just a Github display issue. Same bug in sample output below

local_tests/test_models.py
```diff
@@ -2,16 +2,20 @@
...
 import subprocess
+import traceback
+
+from side_by_side import print_side_by_side
 
 def color_text(s, color):
     return colored.stylize(s, colored.fg(color))
     
-FNULL = open(os.devnull, 'w')
+stderr_file_name = 'err.out.txt'
```
uvafan, line 15: uhhh err.out.txt? Why not just err.out or err.txt?

local_tests/test_models.py
```diff
@@ -2,16 +2,20 @@
...
 
 MAGIC_STRING = '/.*/'
-def compare_outputs(desired_output, test_output):
+def outputs_are_equivalent(desired_output, test_output):
```
uvafan, line 18: what about compare_equivalence? or compare_output_equivalence?

local_tests/test_models.py
```diff
@@ -65,12 +69,22 @@ def log_success(self):
...
-        print('\n')
-        print(f'Test output: {test_output}.')
-        print(f'Correct output: {self.output}.')
+        if errored:
+            print(f'Test exited early with error: {test_output}')
+        else:
+            output1 = f'Test output: {test_output}.'
+            output2 = f'Correct output: {self.output}.'
+            ### begin delete
+            print()
```
uvafan, line 81: condense these 5 lines into one using \n?

local_tests/test_models.py
```diff
@@ -2,16 +2,20 @@
 import io
```
uvafan, line 2: this filename sounds like it's testing the models. Should it be renamed to test_classes or similar? But I guess that could be confusing also

local_tests/test_models.py
```diff
@@ -81,13 +95,23 @@ def __init__(self, command, name=None, output=None, desc=None):
         super().__init__(name=name, output=output, desc=desc)
         
     def execute(self):
+        stderr_file = open(stderr_file_name, 'w+')
         result = subprocess.run(
             self.command.split(), 
             stdout=subprocess.PIPE,
             # @TODO: Collect stderr somewhere. In the event of an error, point user to the error file.
```
uvafan, line 102: Is this TODO done?

local_tests/test_models.py
```diff
@@ -81,13 +95,23 @@ def __init__(self, command, name=None, output=None, desc=None):
...
-            stderr=FNULL 
+            stderr=stderr_file 
         )
-        return result.stdout.decode()
+        stderr_file.seek(0) # go back to beginning of file so we can read the whole thing
+        stderr_str = stderr_file.read()
+        # Remove temp file.
```
uvafan, line 107: Hm is there really no way to capture the stderr into a variable without creating a temp file?

local_tests/test_models.py
```diff
@@ -113,8 +137,26 @@ def __init__(self, function, name=None, output=None, desc=None):
...
+
+def remove_stderr_file():
+    # Make sure the stderr file is removed on exit.
+    try:
+        os.unlink(stderr_file_name)
+    except FileNotFoundError: 
+        # File doesn't exit - that means we never made it or already cleaned it up
+        pass
+    
+def exit_handler(_,__):
```
uvafan, line 158: nit: why is the first arg 1 underscore and the second one 2 haha

textattack/shared/utils.py
```diff
@@ -213,10 +216,4 @@ def html_table_from_rows(rows, title=None, header=None, style_dict=None):
     # Close the table and print to screen.
     table_html += '</table></div>'
     
-    return table_html
-
-def has_letter(word):
```
uvafan, line 218: did you mean to delete this? haha

.travis.yml
```diff
@@ -1,10 +1,9 @@
 language: python
-python: '3.7'
+python: '3.8'
 before_install:
   - python --version
   - pip install -U pip
   - pip install -U pytest
-  - pip install git+https://github.com/jxmorris12/language-check
```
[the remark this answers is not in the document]
jxmorris12: no, it's in `language_tool_python` now

textattack/shared/utils.py
jxmorris12: no, bad merge I guess

local_tests/test_models.py
jxmorris12: I'll leave it for now but I agree there could be  a better name for this and other files in this folder

local_tests/test_models.py
jxmorris12: ok

local_tests/test_models.py
jxmorris12: thanks

local_tests/test_models.py
jxmorris12: good idea thanks

local_tests/command_line_tests.py
```diff
@@ -50,10 +50,26 @@ def register_test(command, name=None, output_file=None, desc=None):
 #                   (takes about 171s)
```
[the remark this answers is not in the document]
jxmorris12: updated

local_tests/command_line_tests.py
jxmorris12: I think we should leave it for now until, until we have tests for specific features, then they can all get shorter names

local_tests/test_models.py
jxmorris12: I tried using `io.StringIO` which would write directly to the buffer but it doesn't work because `subprocess` spawns a whole separate process that writes to a file by its file number and you have to actually open a file for it to have a file number. So, no

PR #91, opened 2020-05-10
minor fixes to install + logging

jxmorris12:
- read cache dir from `os.environ.TA_CACHE_DIR`
- move config.yaml to aws by default; allow user to change it
- add post-install hook so users don't have to run all those commands
- make logger nice

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,0 @@
-include textattack/config.json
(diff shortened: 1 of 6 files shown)


jxmorris12:
Screen Shot 2020-05-09 at 12 50 16 PM

jxmorris12:
Screen Shot 2020-05-09 at 9 50 45 PM

jxmorris12:
need to parse cache_dir from environment instead of getting it from the config.

PR #93, opened 2020-05-13
support swapping multiple words w Augmenter

jxmorris12:


textattack/augmentation/augmenter.py
```diff
@@ -11,40 +15,100 @@ class Augmenter:
             that suggests new texts from an input.
         constraints: (list(textattack.Constraint)): constraints
             that each transformation must meet
+        words_to_swap: (int): Number of words to swap per augmented example
```
uvafan, line 18: So the augmenter only works with word swaps for now?

textattack/augmentation/augmenter.py
```diff
@@ -11,40 +15,100 @@ class Augmenter:
...
-        # Filter out transformations that don't match the constraints.
-        transformations = self._filter_transformations(tokenized_text, transformations)
-        return [t.clean_text() for t in transformations]
+        all_transformations = set()
+        for _ in range(self.transformations_per_example):
+            indices_to_modify = set(range(len(tokenized_text.words)))
+            next_tokenized_text = tokenized_text
+            for __ in range(self.words_to_swap):
+                transformations = []
+                for constraint in self.pre_transformation_constraints:
```
uvafan, line 57: Hmm I see why you need to apply these before calling the transformation the way you have it set up. This way works for everything but `CompositeTransformation`s. Is there a problem with just calling the transformation `self.words_to_swap` times and choosing a random one each time? That way you could use the current method we have for applying pre_modification_constraints and the same index won't be replaced twice since you're enforcing the RepeatModification constraint.

textattack/augmentation/augmenter.py
uvafan: Also maybe name this num_words_to_swap?

textattack/augmentation/augmenter.py
```diff
@@ -11,40 +15,100 @@ class Augmenter:
...
+        """
+        if len(text_list) != len(id_list):
+            raise ValueError('List of text must be same length as list of IDs')
+        if self.transformations_per_example == 0:
+            return text_list, id_list
+        all_text_list = []
+        all_id_list = []
+        if show_progress:
+            text_list = tqdm.tqdm(text_list, desc='Augmenting data...')
+        for text, ids in zip(text_list, id_list):
```
uvafan, line 105: Should this say id? ids makes it seem like it's a list of multiple ids which if I'm understanding it isn't.

textattack/augmentation/augmenter.py
```diff
@@ -11,40 +15,100 @@ class Augmenter:
...
+        """ Supplements a list of text with more text data. Returns the augmented
+            text along with the corresponding IDs for each augmented example.
+        """
+        if len(text_list) != len(id_list):
+            raise ValueError('List of text must be same length as list of IDs')
+        if self.transformations_per_example == 0:
+            return text_list, id_list
+        all_text_list = []
+        all_id_list = []
+        if show_progress:
```
uvafan, line 103: Why a progress bar for this but not for augment_many?

textattack/augmentation/augmenter.py
```diff
@@ -11,40 +15,100 @@ class Augmenter:
...
         `self.transformation`.
+    
+        Args:
+            text_list (list(string)): a list of strings for data augmentation
+            
+        Returns a list(string) of augmented texts.
         """
         return [self.augment(text) for text in text_list]
         
+    def augment_text_and_ids(self, text_list, id_list, show_progress=True):
```
uvafan, line 92: maybe change to augment_text_with_ids or something? this makes it sound like the IDs are being augmented

textattack/augmentation/augmenter.py
uvafan: I guess this would be a bit slower ugh. Another option is to keep what you have now but just try all indices, and if you find an index that gives you 0 transformation options keep choosing random ones until you find one that doesn't. While keeping the break if there are no indices left.

textattack/augmentation/augmenter.py
jxmorris12: fixed and added a comment

textattack/augmentation/augmenter.py
jxmorris12: It's a list of lists of IDs, one list for each tokenized input

textattack/augmentation/augmenter.py
jxmorris12: `augment_many` returns augmentations for a single input, and I can't think of a case where you'd want so many augmentations of a single input that a progress bar is really necessary. This, on the other hand, augments an entire dataset, which can take a lot of time

PR #97, opened 2020-05-15
Add checkpoint feature for running long attacks

jinyongyoo:
Add checkpoint feature so that users can save intermediate results when running attacks that take a long time on a large number of samples. This way, if process crashes for some reason, users can recover their results and continue to run the attack.

Added three command line arguments

- `--checkpoint-resume`: this takes the name of the checkpoint file (`*.ta.chkpt`) that users want to recover from. If set to "latest",  TextAttack automatically chooses latest saved checkpoint from checkpoint directory
- `--checkpoint-dir`: path of the checkpoint directory. This is where TextAttack will save checkpoint files and search for the files when loading. Default is set to `TextAttack/checkpoints`.
- `--checkpoint-interval`: the number of attacks TextAttack should process before saving checkpoint.

`CheckPoint` object is implemented so that is holds all the necessary contents required for recovering from stopped attacks.
Saves:

- Time of when checkpoint was created
- Args: Arguments of the original attack. This is used to recover models, transformations, constraints, etc. This is set once when starting a new run of TextAttack and should not be changed between recoveries. 
- AttackLogManager: Holds the results + information about loggers. Loggers have been modified to be pickleable(?).

### DIFF
--- a/textattack/loggers/visdom_logger.py
+++ b/textattack/loggers/visdom_logger.py
@@ -17,21 +17,19 @@ class VisdomLogger(Logger):
         if not port_is_open(port, hostname=hostname):
             raise socket.error(f'Visdom not running on {hostname}:{port}')
         self.vis = Visdom(port=port, server=hostname, env=env)
-        self.vis_args = {'port': port, 'server': hostname, 'env': env}
+        self.env = env
+        self.port = port
+        self.hostname = hostname
         self.windows = {}
         self.sample_rows = []
 
     def __getstate__(self):
-        # Temporarily save socket handle b/c we can't copy it
-        tmp = self.vis
-        self.vis = None
-        state = copy.deepcopy(self.__dict__)
-        self.vis = tmp
+        state = {i: self.__dict__[i] for i in self.__dict__ if i !='vis'}
         return state
 
     def __setstate__(self, state):
         self.__dict__ = state
-        self.vis = Visdom(**self.vis_args)
+        self.vis = Visdom(port=self.port, server=self.hostname, env=self.env)
 
     def log_attack_result(self, result):
         text_a, text_b = result.diff_color(color_method='html')
--- a/textattack/loggers/visdom_logger.py
+++ b/textattack/loggers/visdom_logger.py
@@ -1,4 +1,5 @@
 import socket
+import copy
 from visdom import Visdom
 
 from textattack.shared.utils import html_table_from_rows
@@ -16,8 +17,21 @@ class VisdomLogger(Logger):
         if not port_is_open(port, hostname=hostname):
             raise socket.error(f'Visdom not running on {hostname}:{port}')
         self.vis = Visdom(port=port, server=hostname, env=env)
+        self.vis_args = {'port': port, 'server': hostname, 'env': env}
         self.windows = {}
         self.sample_rows = []
+
+    def __getstate__(self):
+        # Temporarily save socket handle b/c we can't copy it
+        tmp = self.vis
+        self.vis = None
+        state = copy.deepcopy(self.__dict__)
+        self.vis = tmp
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__ = state
+        self.vis = Visdom(**self.vis_args)
 
     def log_attack_result(self, result):
         text_a, text_b = result.diff_color(color_method='html')
@@ -51,7 +65,7 @@ class VisdomLogger(Logger):
         if not window_id:   window_id = title    # Can provide either of these,
         if not title:       title = window_id    # or both.
         table = html_table_from_rows(rows, title=title, header=header, style_dict=style)
-        self.text(table_html, title=title, window_id=window_id)
+        self.text(table, title=title, window_id=window_id)
 
     def bar(self, X_data, numbins=10, title=None, window_id=None):
         window = None
--- a/textattack/search_methods/attack.py
+++ b/textattack/search_methods/attack.py
@@ -139,6 +139,10 @@ class Attack:
         
         if shuffle:
             random.shuffle(dataset.examples)
+
+        if num_examples <= 0:
+            return
+            yield
             
         for text, ground_truth_output in dataset:
             tokenized_text = TokenizedText(text, self.tokenizer)
--- a/textattack/shared/checkpoint.py
+++ b/textattack/shared/checkpoint.py
@@ -33,17 +33,26 @@ class CheckPoint:
 
         attack_logger_lines = []
         attack_logger_lines.append(utils.add_indent(
-            f'(Number of attacks performed: {self.results_count}', 2
+            f'(Total number of examples to attack): {self.args.num_examples}', 2
         ))
         attack_logger_lines.append(utils.add_indent(
-            f'(Number of successful attacks: {self.num_successful_attacks}', 2
+            f'(Number of attacks performed): {self.results_count}', 2
         ))
         attack_logger_lines.append(utils.add_indent(
-            f'(Number of failed attacks: {self.num_failed_attacks}', 2
+            f'(Number of remaining attacks): {self.num_remaining_attacks}', 2
         ))
-        attack_logger_lines.append(utils.add_indent(
-            f'(Number of skipped attacks: {self.num_skipped_attacks}', 2
+        breakdown_lines = []
+        breakdown_lines.append(utils.add_indent(
+            f'(Number of successful attacks): {self.num_successful_attacks}', 2
         ))
+        breakdown_lines.append(utils.add_indent(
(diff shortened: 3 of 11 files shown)


textattack/shared/checkpoint.py
```diff
@@ -0,0 +1,128 @@
...
+    
+    __str__ = __repr__
+    
+    @property
+    def results_count(self):
+        """ Return number of attacks made so far """
+        return len(self.log_manager.results)
+    
+    @property
+    def num_skipped_attacks(self):
```
jxmorris12, line 71: nit: you could condense each of these 3 methods into something like:


```
def num_skipped_attacks(self):
   return len([r for r in self.log_manager.results if isinstance(r, SkippedAttackResult])
```

or even more python-y:

def num_skipped_attacks(self):
   return sum(isinstance(r, SkippedAttackResult) for r in self.log_manager.results)
```

(this last one works because the sum of a bunch of `bool`s is equal to the number of true values)

File: textattack/search_methods/attack.py [hunk 2]
```diff
@@ -139,6 +139,10 @@ def _get_examples_from_dataset(self, dataset, num_examples=None, shuffle=False,
         
         if shuffle:
             random.shuffle(dataset.examples)
+
+        if num_examples <= 0:
+            return
+            yield
```
jxmorris12, line 145: I don't think this `yield` is supposed to be here

textattack/loggers/visdom_logger.py
```diff
@@ -16,9 +17,22 @@ def __init__(self, env='main', port=8097, hostname='localhost'):
...
+        # Temporarily save socket handle b/c we can't copy it
+        tmp = self.vis
+        self.vis = None
+        state = copy.deepcopy(self.__dict__)
+        self.vis = tmp
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__ = state
+        self.vis = Visdom(**self.vis_args)
```
jxmorris12, line 34: Consider removing self.vis_args and saving the other args as variables, then changing this line to 
`self.vis = Visdom(port=self.port, server=self.hostname, env=self.env)`

textattack/loggers/visdom_logger.py
```diff
@@ -16,9 +17,22 @@ def __init__(self, env='main', port=8097, hostname='localhost'):
...
         self.sample_rows = []
 
+    def __getstate__(self):
+        # Temporarily save socket handle b/c we can't copy it
+        tmp = self.vis
+        self.vis = None
+        state = copy.deepcopy(self.__dict__)
+        self.vis = tmp
+        return state
+
```
jxmorris12, line 31: I see this pattern here and `FileLogger` and was trying to think of a better way. Here's another option, using a dictionary comprehension:

`state = {i: d[i] for i in self.__dict__ if i !='vis'}`

not saying this is necessarily better, it's up to you whether you'd want to change the two `__getstate__` methods to look like this. It's probably a little less easy to comprehend this way but a decent amount more concise. Up to you.

textattack/shared/checkpoint.py
```diff
@@ -0,0 +1,128 @@
...
+            time (float): epoch time representing when checkpoint was made
+            args: command line arguments of the original attack
+            log_manager (AttackLogManager)
+    """
+    def __init__(self, time, args, log_manager):
+        self.time = time
+        self.args = args
+        self.log_manager = log_manager
+
+    def __repr__(self):
```
jxmorris12, line 20: great use of __repr__

textattack/shared/scripts/run_attack_parallel.py
```diff
@@ -94,6 +113,16 @@ def run(args):
...
             in_queue.put((label, text))
+
+        if args.checkpoint_interval and num_results % args.checkpoint_interval == 0:
+            chkpt_time = time.time()
+            date_time = datetime.datetime.fromtimestamp(chkpt_time).strftime('%Y-%m-%d %H:%M:%S')
+            print('\n\n' + '=' * 100)
+            logger.info('Saving checkpoint at {} after {} attacks.'.format(date_time, num_results))
+            print('=' * 100 + '\n')
+            checkpoint = textattack.shared.CheckPoint(chkpt_time, args, attack_log_manager)
+            checkpoint.save()
```
jxmorris12, line 124: I see this code repeated here and in `run_attack_single_threaded`. Maybe you could add the `CheckPoint` (now `Checkpoint`) class somehow? Then you could call 

``` 
checkpoint = ...
checkpoint.save(quiet=False)
```
and it'd automatically print that stuff. just an idea.

textattack/shared/checkpoint.py
```diff
@@ -0,0 +1,128 @@
...
+
+class CheckPoint:
+    """ An object that stores necessary information for saving and loading checkpoints
+    
+        Args:
+            time (float): epoch time representing when checkpoint was made
+            args: command line arguments of the original attack
+            log_manager (AttackLogManager)
+    """
+    def __init__(self, time, args, log_manager):
```
jxmorris12, line 15: maybe `time` should be optional and should default to `time.time()`?

textattack/shared/scripts/run_attack_args_helper.py
```diff
@@ -335,3 +349,24 @@ def parse_logger_from_args(args):# Create logger
     if not args.disable_stdout:
         attack_log_manager.enable_stdout()
     return attack_log_manager
+
+def parse_checkpoint_from_args(args):
+    if args.checkpoint_resume:
+        if args.checkpoint_resume.lower() == 'latest':
+            chkpt_files = [f for f in os.listdir(args.checkpoint_dir) if f.endswith('.ta.chkpt')]
+            latest_file = max(chkpt_files)
```
jxmorris12, line 357: are you sure this works how you intend it to? what about this:

```
chkpt_files = ['check_9.chkpt', 'check_10.chkpt']
max(chkpt_files)
```

produce `check_9` instead of `check_10`?

jinyongyoo: Default format is `{time}.ta.chkpt`, so `max` should select the latest item. The way it is set up right now is that users cannot set the name of checkpoint files.

textattack/shared/checkpoint.py
jinyongyoo: Good idea

textattack/search_methods/attack.py
jinyongyoo: I included `yield` after seeing this: https://stackoverflow.com/questions/13243766/python-empty-generator-function

I'm not familiar with generators, so I'm not sure which is correct.

textattack/loggers/visdom_logger.py
jinyongyoo: Done

textattack/shared/checkpoint.py
jinyongyoo: Good point

textattack/loggers/visdom_logger.py
jinyongyoo: I agree that this is better.

textattack/shared/scripts/run_attack_parallel.py
jinyongyoo: I wrote the prints msgs before coming up with `Checkpoint`. Good idea

textattack/shared/scripts/run_attack_args_helper.py
```diff
@@ -196,6 +197,15 @@ def get_args():
     
     def str_to_int(s): return sum((ord(c) for c in s))
     parser.add_argument('--random-seed', default=str_to_int('TEXTATTACK'))
+
+    parser.add_argument('--checkpoint-resume', required=False, type=str,
```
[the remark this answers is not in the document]
jinyongyoo: To achieve this, I'm assuming that we would group all the regular arguments into an argument group and make that argument group mutually exclusive with `--checkpoint-resume`. But within that arg group, we need another mutually exclusive group for `--attack` and `--recipe`. I'm not sure if nested argument groups and nested mutually exclusive groups are supported in Python...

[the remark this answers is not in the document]
jinyongyoo: Also turns out grouping arguments are just for sake of displaying help messages. So there's no way of making `--checkpoint-resume` vs `everything else` possible using mutually exclusive groups.

I changed it by making `resume` a sub-command so that we process arguments differently.

textattack/search_methods/attack.py
jxmorris12: Whoa, that's weird. I think `return []` would work fine here but the generator syntax is good too! We can leave it.

PR #113, opened 2020-06-04
Improved search space

jxmorris12:
- Change `TokenizedText` to track insertions & deletions, allowing for a larger space of attacks
- Add `WordDeletion` transformation
- add Input Reduction attack from [Pathologies of Neural Models Make Interpretations Difficult](https://arxiv.org/abs/1804.07781)
- add [TextBugger](https://arxiv.org/abs/1812.05271)
- WordSwapHomoglyph --> WordSwapHomoglyphSwap (I know it sounds dumb when I right it but it's to main consistency with the other char-swap transformations)
- allow Greedy-WIR to rank things in order of _ascending_ importance
- fix error that occurs with fast_tokenizer on certain versions of `tokenizer`
- update test scripts to add `--quiet`


textattack/attack_recipes/input_reduction_feng_2018.py
```diff
@@ -0,0 +1,42 @@
...
+    # "For each word in an input sentence, we measure its importance by the 
+    # change in the confidence of the original prediction when we remove
+    # that word from the sentence."
+    #
+    # "Instead of looking at the words with high importance values—what 
+    # interpretation methods commonly do—we take a complementary approach
+    # and study how the model behaves when the supposedly unimportant words are 
+    # removed."
+    #
+    search_method = GreedyWordSwapWIR(wir_method='delete', ascending=True)
```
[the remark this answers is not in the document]
jinyongyoo: Yeah I agree. Maybe we can come up with a new goal_function for this specific task?

textattack/shared/tokenized_text.py
```diff
@@ -155,21 +180,59 @@ def replace_new_words(self, new_words):
...
-            if i in self.attack_attrs['modified_indices'] or input_word != adv_word:
-                new_attack_attrs['modified_indices'].add(new_i)
+            adv_num_words = len(words_from_text(adv_word))
+            num_words_diff = adv_num_words - len(words_from_text(input_word))
+            # Track indices on insertions and deletions.
+            if num_words_diff != 0:
+                # Re-calculated modified indices. If words are inserted or deleted, 
+                # they could change.
+                shifted_modified_indices = set()
+                for j, modified_idx in enumerate(new_attack_attrs['modified_indices']):
```
uvafan, line 200: j isn't used?

textattack/transformations/transformation.py
```diff
@@ -25,6 +25,7 @@ def __call__(self, current_text, pre_transformation_constraints=[], indices_to_m
             indices_to_modify = set(indices_to_modify)
         for constraint in pre_transformation_constraints:
             indices_to_modify = indices_to_modify & constraint(current_text, self)
+        indices_to_modify = current_text.convert_from_original_idxs(indices_to_modify)
```
uvafan, line 28: I think this method should assume `indices_to_modify` are up-to-date indices unless indicated otherwise, and have a boolean flag to indicate they're not. And this conversion definitely be above the for loop because the constraints act on the current indices not the original ones.

uvafan:
A few ideas on how to refactor this insertion/deletion stuff:
Less speculative:
- I personally think `WordDeletion` shouldn't be a subclass of `WordSwap`, since it's not a word swap. I also think the one that splits a words into multiple words should be implemented separate from `WordSwap`. If you want you can rename `WordSwap` to `WordModification` maybe? Then create subclasses from there? But imo these aren't swaps.
- Relatedly, I think the deletion of words and splitting up of a word should probably be handled by `TokenizedText` with methods like `delete_word_at_idx` or `split_up_word_at_idx`. And in the future `insert_word_at_idx`. This seems more intuitive and easier to understand. Even if these methods just call `replace_words_at_indices` for now.

More speculative:
- What if we maintain 2 dicts: (A) mapping from original indices to current indices and (B) mapping from current indices to original indices? State transitions shouldn't be too hard to work out, and I feel like this could cut down on the complexity of the code (less loops to figure out shifts etc. when need to know a shift can just look at the dict). Also these state transitions could be handled in the methods suggested above (`delete_word_at_idx` etc.). This would also remove the need for two separate lists `modified_indices` and `original_modified_indices`. 
- Armed with the (A) dict `convert_from_original_idxs` would be a one liner.

textattack/attack_recipes/textbugger_li_2018.py
```diff
@@ -0,0 +1,69 @@
...
+        StopwordModification()
+    ]
+    # In our experiment, we first use the Universal Sentence
+    # Encoder [7], a model trained on a number of natural language
+    # prediction tasks that require modeling the meaning of word
+    # sequences, to encode sentences into high dimensional vectors.
+    # Then, we use the cosine similarity to measure the semantic
+    # similarity between original texts and adversarial texts
+    constraints.append(
+        UniversalSentenceEncoder()
```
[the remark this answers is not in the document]
jxmorris12: found it:
```
"Furthermore, the semantic similarity threshold is set as 0.8 to guarantee a good trade-off between quality and strength of the generated adversarial text."
```

docs/attacks/attack_recipes.rst
```diff
@@ -40,3 +40,17 @@ Seq2Sick
...
+
+
+TextFooler
+###########
+
+.. automodule:: textattack.attack_recipes.textfooler_jin_2019
+   :members:
+
+
+TextBuffer
```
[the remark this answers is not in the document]
jxmorris12: lol one keystroke away

textattack/attack_recipes/input_reduction_feng_2018.py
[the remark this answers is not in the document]
jxmorris12: Ok, removed for now.

textattack/shared/tokenized_text.py
```diff
@@ -155,21 +180,59 @@ def replace_new_words(self, new_words):
         final_sentence = ''
         text = self.text
         new_attack_attrs = dict()
-        new_attack_attrs['modified_indices'] = set()
+        new_attack_attrs['original_modified_indices'] = self.attack_attrs['original_modified_indices'].copy()
+        new_attack_attrs['modified_indices'] = self.attack_attrs['modified_indices'].copy()
         new_attack_attrs['newly_modified_indices'] = set()
         new_i = 0
         for i, (input_word, adv_word) in enumerate(zip(self.words, new_words)):
             if input_word == '[DELETE]': continue
```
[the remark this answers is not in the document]
jxmorris12: honestly not 100% sure why it was ever here!

textattack/shared/tokenized_text.py
```diff
@@ -155,21 +180,59 @@ def replace_new_words(self, new_words):
...
+                for j, modified_idx in enumerate(new_attack_attrs['modified_indices']):
+                    if modified_idx < i:
+                        shifted_modified_indices.add(modified_idx)
+                    elif modified_idx > i:
+                        shifted_modified_indices.add(modified_idx + num_words_diff)
+                    else:
+                        pass
+                new_attack_attrs['modified_indices'] = shifted_modified_indices
+                # Track insertions and deletions wrt original text.
+                if num_words_diff != 0:
```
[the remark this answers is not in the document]
jxmorris12: oops

textattack/transformations/transformation.py
jxmorris12: yeah, the boolean is at the beginning of `tt.convert_from_original_idxs`. So it doesn't do anything expensive unless `len(tt.attack_attrs['original_modified_indices']) > 0`. Good call about moving the conversion up

jxmorris12:
High-level updates:

- changed `WordDeletion` to just subclass `Transformation`. It's a few extra lines of code but, as you pointed out, probably a useful semantic distinction
- removed input reduction recipe [until maximization goal functions exist]
- changed `TokenizedText` to map indices from original -> perturbed as an nparray: `original_index_map`
- added option to ignore first/last char for TextBugger transformations

textattack/transformations/transformation.py
uvafan: What boolean?

jxmorris12: https://github.com/QData/TextAttack/blob/ec857cb59f3233479747ecc9519eae363920bf28/textattack/shared/tokenized_text.py#L159

uvafan: You mean when there's no `original_index_map`? I'm trying to say that the default behavior should be to assume that the argument `indices_to_modify` represents the current rather than original indices. In our case our current things pass in original indices but I don't think that makes sense as the default. So it has nothing to do with whether the TT has an `original_index_map` yet.

jxmorris12: In the default case (no shifts) `original_index_map` won't exist so convert_from_original_idxs will return the same object that was passed in. I don't see anywhere where the original indices are passed in.

textattack/transformations/transformation.py
```diff
@@ -18,11 +18,17 @@ def __call__(self, current_text, pre_transformation_constraints=[], indices_to_m
                 beginning the transformation.
             indices_to_modify: Which word indices should be modified as dictated by the
                 ``SearchMethod``.
+            shifted_idxs (bool): Whether indices have been shifted from
```
uvafan, line 21: This is kinda unclear. I think it should be named `original_idxs` and say something like "Whether indices represent those from the original text rather than the current one."

textattack/transformations/word_deletion.py
```diff
@@ -0,0 +1,18 @@
...
+
+from .transformation import Transformation
+
+class WordDeletion(Transformation):
+    """
+    An abstract class that takes a sentence and transforms it by replacing
+    deleting a single word.
+    
+    Args:
+        letters_to_insert (string): letters allowed for insertion into words
```
uvafan, line 11: remove args from docstring

textattack/attack_recipes/textbugger_li_2018.py
```diff
@@ -0,0 +1,72 @@
...
+    ArXiv, abs/1812.05271.
+    """
+    #
+    #  we propose five bug generation methods for TEXTBUGGER:
+    #
+    transformation = CompositeTransformation([
+        # (1) Insert: Insert a space into the word.
+        # Generally, words are segmented by spaces in English. Therefore,
+        # we can deceive classifiers by inserting spaces into words.
+        WordSwapRandomCharacterInsertion(random_one=True, letters_to_insert=' ', skip_first_char=True, skip_last_char=True),
```
uvafan, line 27: I still think we should make a separate transformation for splitting up words. It's not a word swap. If you prefer it this way we should discuss it.

textattack/transformations/word_swap_neighboring_character_swap.py
```diff
@@ -9,10 +9,16 @@ class WordSwapNeighboringCharacterSwap(WordSwap):
         Args:
             random_one (bool): Whether to return a single word with two characters 
                 swapped. If not, returns all possible options.
+            skip_first_char (bool): Whether to disregard perturbing the first 
+                character.
+            skip_last_char (bool): Whether to disregard perturbing the last 
+                character.
     """
-    def __init__(self, random_one=True, **kwargs):
+    def __init__(self, random_one=True, skip_first_char=False, skip_last_char=False, **kwargs):
```
uvafan, line 17: nit: Is there a reason we have separate skip_first_char and skip_last_char args rather than a replace_ends bool?

textattack/transformations/word_swap_random_character_deletion.py
```diff
@@ -22,13 +28,16 @@ def _get_replacement_words(self, word):
             return []
             
         candidate_words = []
+        
+        start_idx = 1 if self.skip_first_char else 0
+        end_idx = (len(word) - 13) if self.skip_last_char else len(word)
```
uvafan, line 33: typo. -13->-1

textattack_outputs/attack-1591504828757.txt
```diff
@@ -0,0 +1,42 @@
+--------------------------------------------- Result 1 ---------------------------------------------
```
uvafan, line 1: Need to gitignore the `textattack_outputs` folder, or remove it from commit

textattack/transformations/word_swap_random_character_deletion.py
jxmorris12: lol oops

PR #128, opened 2020-06-16
Worklist addition

jinyongyoo:
The issue with checkpoints right now is that when running in parallel mode, results are logged asynchronously, which means we might have logged results 8, 10, and 11, but not result 9 b/c the attack is taking longer. Now in the case that we make a checkpoint and crash before result 9 can finish, what happens is that the next time we recover, we start processing from 12th example and entirely ignore the 9th example that was not logged. This is because we manipulate dataset offset, which (1) isn't nice, (2) doesn't work with `nlp` dataset.

So, we eschew from manipulating the dataset offset and rely on maintaining a `worklist`, which is a `deque` object that stores all the indicies of examples we still need to attack with respect to the dataset. This is easy to adapt for parallel mode, but is more tricky for single-thread version because we rely on generators. I did make it work and have performed some testing to ensure it works. Now, the bad news is that it's hard to write tests for checkpoints since we have to imitate crash behaviors, which isn't easy with testing frameworks. I also think there's a lot of ways this can be improved, so any ideas/critiques are welcomed.

On another note, removed `shuffle` operation from `attack` and instead made it part of `__init__` of datasets. It makes more sense that we should be shuffle when we first load the data, not right before we attack it.


textattack/shared/scripts/run_attack_parallel.py
```diff
@@ -104,28 +109,39 @@ def run(args):
         num_results = 0
         num_failures = 0
         num_successes = 0
-    pbar = tqdm.tqdm(total=num_remaining_examples, smoothing=0)
-    while num_results < num_total_examples:
+    pbar = tqdm.tqdm(total=num_remaining_attacks, smoothing=0)
+    while num_results < num_total_examples and worklist:
```
jxmorris12, line 113: can this just say `while worklist`?

textattack/shared/scripts/run_attack_parallel.py
```diff
@@ -48,15 +49,19 @@ def run(args):
         # Override current args with checkpoint args
         resume_checkpoint = parse_checkpoint_from_args(args)
         args = merge_checkpoint_args(resume_checkpoint.args, args)
-        num_examples_offset = resume_checkpoint.dataset_offset
-        num_remaining_examples = resume_checkpoint.num_remaining_attacks
+        
+        num_remaining_attacks = resume_checkpoint.num_remaining_attacks
         num_total_examples = args.num_examples
+        worklist = resume_checkpoint.worklist.copy()
```
jxmorris12, line 55: do you need to copy here? jw

textattack/shared/scripts/run_attack_parallel.py
```diff
@@ -104,28 +109,39 @@ def run(args):
...
+                worklist.append(last_example)
+                in_queue.put((last_example, label, text))
+            except IndexError:
+                last_example -= 1
 
         if args.checkpoint_interval and num_results % args.checkpoint_interval == 0:
             attack_log_manager.flush()
-            checkpoint = textattack.shared.Checkpoint(args, attack_log_manager)
+            checkpoint = textattack.shared.Checkpoint(args, attack_log_manager, worklist, last_example)
+            if not checkpoint.verify_no_duplicates():
```
jxmorris12, line 143: Okay, I have some questions about this.

1. Why are duplicate results allowed? (Why do you warn instead of erroring out?)

2. Why isn't this check completed in the `Checkpoint` constructor?

3. How can this ever happen?

textattack/shared/scripts/run_attack_single_threaded.py
```diff
@@ -27,13 +28,18 @@ def run(args):
...
+        last_example = resume_checkpoint.last_example
+
         logger.info('Recovered from checkpoint previously saved at {}'.format(resume_checkpoint.datetime))
         print(resume_checkpoint, '\n')
     else:
-        num_examples_offset = args.num_examples_offset
-        num_examples = args.num_examples
+        num_remaining_attacks = args.num_examples
+        worklist = deque(range(0, args.num_examples))
+        last_example = worklist[-1]
```
jxmorris12, line 42: I think I understand what `last_example` is doing, but can you rename it (maybe to `last_example_idx` or something else clearer) and add a comment or two clarifying that it's necessary to enforce `--attack-n`?

textattack/shared/scripts/run_attack_parallel.py
```diff
@@ -104,28 +109,39 @@ def run(args):
...
+
             if type(result) == textattack.attack_results.SuccessfulAttackResult:
                 num_successes += 1
             if type(result) == textattack.attack_results.FailedAttackResult:
                 num_failures += 1
             pbar.set_description('[Succeeded / Failed / Total] {} / {} / {}'.format(num_successes, num_failures, num_results))
         else:
-            label, text = next(dataset)
-            in_queue.put((label, text))
+            last_example += 1
```
jxmorris12, line 132: same comments below about `last_example`

textattack/datasets/classification/ag_news.py
```diff
@@ -33,6 +33,8 @@ class AGNews(ClassificationDataset):
     
     """
     DATA_PATH = 'datasets/classification/ag_news.txt'
-    def __init__(self, offset=0):
+    def __init__(self, offset=0, shuffle=False):
```
jxmorris12, line 36: please add `shuffle` to docstring and remove `n`

textattack/datasets/classification/imdb_sentiment.py
```diff
@@ -14,6 +14,8 @@ class IMDBSentiment(ClassificationDataset):
     
     """
     DATA_PATH = 'datasets/classification/imdb.txt'
-    def __init__(self, offset=0):
+    def __init__(self, offset=0, shuffle=False):
```
jxmorris12, line 17: please add `shuffle` to docstring and remove `n`

textattack/datasets/classification/kaggle_fake_news.py
```diff
@@ -14,6 +14,8 @@ class KaggleFakeNews(ClassificationDataset):
     
     """
     DATA_PATH = 'datasets/classification/fake'
-    def __init__(self, offset=0):
+    def __init__(self, offset=0, shuffle=False):
```
jxmorris12, line 17: please add `shuffle` to docstring and remove `n`

textattack/datasets/classification/movie_review_sentiment.py
```diff
@@ -16,6 +16,8 @@ class MovieReviewSentiment(ClassificationDataset):
     
     """
     DATA_PATH = 'datasets/classification/mr.txt'
-    def __init__(self, offset=0):
+    def __init__(self, offset=0, shuffle=False):
```
jxmorris12, line 19: please add `shuffle` to docstring and remove `n` (sorry)

textattack/datasets/classification/yelp_sentiment.py
```diff
@@ -13,9 +13,11 @@ class YelpSentiment(ClassificationDataset):
     
     """
     DATA_PATH = 'datasets/classification/yelp_sentiment.txt'
-    def __init__(self, offset=0):
+    def __init__(self, offset=0, shuffle=False):
```
jxmorris12, line 16: please add `shuffle` to docstring

textattack/datasets/entailment/mnli.py
```diff
@@ -18,7 +18,9 @@ class MNLI(EntailmentDataset):
     """
     MATCHED_DATA_PATH = 'datasets/entailment/mnli_matched'
     MISMATCHED_DATA_PATH = 'datasets/entailment/mnli_mismatched'
-    def __init__(self, offset=0, mismatched=False):
+    def __init__(self, offset=0, mismatched=False, shuffle=False):
```
jxmorris12, line 21: please add `shuffle` to docstring

textattack/datasets/entailment/snli.py
```diff
@@ -14,6 +14,8 @@ class SNLI(EntailmentDataset):
     
     """
     DATA_PATH = 'datasets/entailment/snli'
-    def __init__(self, offset=0):
+    def __init__(self, offset=0, shuffle=False):
```
jxmorris12, line 17: please add `shuffle` to docstring

textattack/datasets/translation/translation_datasets.py
```diff
@@ -11,5 +11,7 @@ class NewsTest2013EnglishToGerman(TextAttackDataset):
     
     """
     DATA_PATH = 'datasets/translation/NewsTest2013EnglishToGerman'
-    def __init__(self, offset=0):
+    def __init__(self, offset=0, shuffle=False):
```
jxmorris12, line 14: please add `shuffle` and `offset` to docstring

textattack/shared/scripts/run_attack_parallel.py
```diff
@@ -104,28 +109,39 @@ def run(args):
...
             pbar.update()
             num_results += 1
+
             if type(result) == textattack.attack_results.SuccessfulAttackResult:
                 num_successes += 1
             if type(result) == textattack.attack_results.FailedAttackResult:
                 num_failures += 1
             pbar.set_description('[Succeeded / Failed / Total] {} / {} / {}'.format(num_successes, num_failures, num_results))
         else:
-            label, text = next(dataset)
```
jxmorris12, line 123: I think this is a good change. Can you remove the `._i` indexing from the `TextAttackDataset` altogether, and remove the `__next__` method from there?

jxmorris12:
this is a really good change. Almost all of my comments are documentation-related. Thanks for making datasets nicer and moving `shuffle` into the dataset class.

textattack/datasets/classification/ag_news.py
```diff
@@ -33,6 +33,8 @@ class AGNews(ClassificationDataset):
...
         """ Loads a full dataset from disk. """
         self._load_classification_text_file(AGNews.DATA_PATH, offset=offset)
+        if shuffle:
```
jxmorris12, line 39: Two questions about this `shuffle` pattern:

1. can we move all the shuffle checks into `_load_classification_text_file` so the code isn't repeated?

2. will this shuffling work with `nlp` datasets?

textattack/shared/scripts/run_attack_parallel.py
jinyongyoo: Probably not.

textattack/shared/scripts/run_attack_parallel.py
jinyongyoo: yes

textattack/shared/attack.py
```diff
@@ -142,15 +141,14 @@ def attack_one(self, initial_result):
         else:
             return FailedAttackResult(initial_result, final_result, self.goal_function.num_queries)
  
-    def _get_examples_from_dataset(self, dataset, num_examples=None, shuffle=False,
+    def _get_examples_from_dataset(self, dataset, indicies=None,
```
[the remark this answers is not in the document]
jinyongyoo: good catch

textattack/shared/scripts/run_attack_single_threaded.py
jinyongyoo: Yeah it needs a better name.

textattack/datasets/classification/ag_news.py
jinyongyoo: 1. Done.
2. NLP dataset class already handles shuffling the same way.

textattack/datasets/entailment/snli.py
jinyongyoo: done

textattack/datasets/translation/translation_datasets.py
jinyongyoo: done

textattack/datasets/entailment/mnli.py
jinyongyoo: done

textattack/datasets/classification/yelp_sentiment.py
jinyongyoo: done

textattack/shared/scripts/run_attack_parallel.py
jinyongyoo: I thought maybe it would be better to allow users to have duplicates and have them recover it. But now I think about it, it's highly unlikely and if it happens, i agree we should just error out.

Issue #126, opened 2020-06-13
Remove `ids` from `TokenizedText`


jinyongyoo:
It seems like the only use of `ids` for `TokenizedText` object is for using it when calling the victim model in `goal_function`. This leads to unnecessary tokenization operations when we're generating `TokenizedText` from transformations since some of them will get filtered out by the constraints and won't be passed to the victim model. 

It also adds confusion as to what the model `ids` are for for newcomers. For instance, they might wonder if it's for the victim model or one of the language models. I think we should remove the `id` tokenization from `TokenizedText` and have `goal_function` tokenize the text.

PR #129, opened 2020-06-17
Tokenized text refactor

jxmorris12:
- `TokenizedText` -> `AttackedText`
- pass `OrderedDict` objects from datasets so that the keys stay with them (like 'premise' and 'hypothesis'
- print those keys in `AttackedText`
- remove tokens from `AttackedText`

### DIFF
--- a/tests/sample_outputs/run_attack_deepwordbug_bert_snli_10.txt
+++ b/tests/sample_outputs/run_attack_deepwordbug_bert_snli_10.txt
@@ -16,29 +16,29 @@ Attack(
 Load time: /.*/
 --------------------------------------------- Result 1 ---------------------------------------------
 [91m0[0m-->[37m[SKIPPED][0m
-A person in a black and green outfit is riding a bicycle .
+Premise: A person in a black and green outfit is riding a bicycle .
 
-a colorfully dressed child rides a bike
+Hypothesis: a colorfully dressed child rides a bike
 
 
 --------------------------------------------- Result 2 ---------------------------------------------
 [92m1[0m-->[91m0[0m
-A person rolls down a hill riding a wagon as another watches .
+Premise: A person rolls down a hill riding a wagon as another watches .
 
-A [92mchild[0m in a wagon rolls down a hill .
-A person rolls down a hill riding a wagon as another watches .
+Hypothesis: A [92mchild[0m in a wagon rolls down a hill .
+Premise: A person rolls down a hill riding a wagon as another watches .
 
-A [91mcihld[0m in a wagon rolls down a hill .
+Hypothesis: A [91mcihld[0m in a wagon rolls down a hill .
 
 
 --------------------------------------------- Result 3 ---------------------------------------------
 [94m2[0m-->[92m1[0m
-A man in a black tank top wearing a red plaid [94mhat[0m
+Premise: A man in a black tank top wearing a red plaid [94mhat[0m
 
-A man wearing football pads .
-A man in a black tank top wearing a red plaid [92mat[0m
+Hypothesis: A man wearing football pads .
+Premise: A man in a black tank top wearing a red plaid [92mat[0m
 
-A man wearing football pads .
+Hypothesis: A man wearing football pads .
 
 
 
--- a/tests/sample_outputs/run_attack_deepwordbug_bert_snli_10.txt
+++ b/tests/sample_outputs/run_attack_deepwordbug_bert_snli_10.txt
@@ -16,29 +16,29 @@ Attack(
 Load time: /.*/
 --------------------------------------------- Result 1 ---------------------------------------------
 [91m0[0m-->[37m[SKIPPED][0m
-Premise: A person in a black and green outfit is riding a bicycle .
 
-Hypothesis: a colorfully dressed child rides a bike
+[1m[4mPremise[0m[0m: A person in a black and green outfit is riding a bicycle .
+[1m[4mHypothesis[0m[0m: a colorfully dressed child rides a bike
 
 
 --------------------------------------------- Result 2 ---------------------------------------------
 [92m1[0m-->[91m0[0m
-Premise: A person rolls down a hill riding a wagon as another watches .
 
-Hypothesis: A [92mchild[0m in a wagon rolls down a hill .
-Premise: A person rolls down a hill riding a wagon as another watches .
+[1m[4mPremise[0m[0m: A person rolls down a hill riding a wagon as another watches .
+[1m[4mHypothesis[0m[0m: A [92mchild[0m in a wagon rolls down a hill .
 
-Hypothesis: A [91mcihld[0m in a wagon rolls down a hill .
+[1m[4mPremise[0m[0m: A person rolls down a hill riding a wagon as another watches .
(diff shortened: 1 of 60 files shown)


tests/sample_outputs/run_attack_deepwordbug_bert_snli_10.txt
```diff
@@ -16,29 +16,29 @@ Attack(
...
+Premise: A person rolls down a hill riding a wagon as another watches .
 
-A [91mcihld[0m in a wagon rolls down a hill .
+Hypothesis: A [91mcihld[0m in a wagon rolls down a hill .
 
 
 --------------------------------------------- Result 3 ---------------------------------------------
 [94m2[0m-->[92m1[0m
-A man in a black tank top wearing a red plaid [94mhat[0m
+Premise: A man in a black tank top wearing a red plaid [94mhat[0m
```
uvafan, line 36: Wouldn't it make more sense to have a blank line between Premise-Hypothesis pairs than between each premise and hypothesis?

textattack/models/tokenizers/auto_tokenizer.py
```diff
@@ -29,12 +30,28 @@ def __init__(
...
+        """ The batch equivalent of ``encode``."""
+        if hasattr(self.tokenizer, "encode_batch"):
+            return self.tokenizer.encode_batch(
+                input_text_list,
+                max_length=self.max_length,
+                add_special_tokens=True,
+                pad_to_max_length=True,
+            )
+        else:
+            return [dict(self.encode(input_text)) for input_text in input_text_list]
```
uvafan, line 57: Doesn't self.encode(input_text) already return a dict?

textattack/shared/attack.py
```diff
@@ -182,85 +182,64 @@ def attack_one(self, initial_result):
...
+            i = indices.popleft()
+            try:
+                text, ground_truth_output = dataset[i]
+                tokenized_text = AttackedText(text)
+                self.goal_function.num_queries = 0
+                goal_function_result, _ = self.goal_function.get_result(
+                    tokenized_text, ground_truth_output
+                )
+                if goal_function_result.succeeded:
+                    # Store the true output on the goal function so that the
```
uvafan, line 213: Apparently this was already here but this is confusing to me. Why do we have to set the output? Feels like there should be a better way to do this. Maybe I can change it as part of my score function PR?

textattack/shared/attacked_text.py
```diff
@@ -77,16 +96,16 @@ def free_memory(self):
     def text_window_around_index(self, index, window_size):
         """ The text window of ``window_size`` words centered around ``index``. """
         length = len(self.words)
-        half_size = (window_size - 1) // 2
+        half_size = (window_size - 1) / 2.0
         if index - half_size < 0:
             start = 0
-            end = min(window_size, length - 1)
-        elif index + half_size > length - 1:
+            end = min(window_size - 1, length - 1)
```
uvafan, line 102: good catch

textattack/shared/attacked_text.py
```diff
@@ -231,24 +282,28 @@ def replace_new_words(self, new_words):
             with multiple space-separated words, representation an insertion
             of one or more words.
         """
-        final_sentence = ""
-        text = self.text
+        perturbed_text = ""
+        original_text = AttackedText.SPLIT_TOKEN.join(self._text_input.values())
```
uvafan, line 286: did you test inserting at the end of the premise and beginning of hypothesis? feels like the split token could mess this up

uvafan: Also maybe current_text is a better tname than original_text?

uvafan: It feels weird that we still have to use the SPLIT_TOKEN here

uvafan:
Closes https://github.com/QData/TextAttack/issues/126

textattack/shared/attack.py
jinyongyoo: Well that's for when model incorrectly predicts for the given output, so we display ground truth for skipped samples. I'm not sure if there's a better place to being doing this since it's the one place where we're handling `SkippedAttackResult`

textattack/shared/attacked_text.py
```diff
@@ -64,10 +83,10 @@ def __hash__(self):
     def free_memory(self):
         """ Delete items that take up memory.
             
-            Can be called once the TokenizedText is only needed to display.
+            Can be called once the AttackedText is only needed to display.
         """
-        self.ids = None
-        self.tokenizer = None
+        if "previous_attacked_text" in self.attack_attrs:
+            self.attack_attrs["previous_attacked_text"].free_memory()
```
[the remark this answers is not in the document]
jxmorris12: Well it's just a pointer to the previous text, so I think it makes more sense to keep it. My idea was because I thought there should be a format where we print all the steps of a perturbation from the original to perturbed

textattack/shared/attacked_text.py
```diff
@@ -77,16 +96,16 @@ def free_memory(self):
...
-            end = min(window_size, length - 1)
-        elif index + half_size > length - 1:
+            end = min(window_size - 1, length - 1)
+        elif index + half_size >= length:
             start = max(0, length - window_size)
             end = length - 1
         else:
-            start = index - half_size
-            end = index + half_size
+            start = index - math.ceil(half_size)
```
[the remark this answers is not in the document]
jxmorris12: because if window length is odd, we have to "choose" whether to return an extra word from one side or the other

tests/sample_outputs/run_attack_deepwordbug_bert_snli_10.txt
jxmorris12: yes

textattack/shared/attack.py
jxmorris12: yes I agree @uvafan you should change this – every `AttackedText` object should have its ground truth output stored somewhere I think. Then we can print it to the CSV no matter what.

textattack/shared/attacked_text.py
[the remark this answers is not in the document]
jinyongyoo: Hmm yeah that might be useful for the demo.

textattack/shared/attacked_text.py
```diff
@@ -211,17 +233,46 @@ def replace_word_at_index(self, index, new_word):
...
+        # optionally capitalize ``text``.
+        new_text = " ".join((text, word_at_index))
+        return self.replace_word_at_index(index, new_text)
 
     def get_deletion_indices(self):
         return self.attack_attrs["original_index_map"][
             self.attack_attrs["original_index_map"] == -1
         ]
 
     def replace_new_words(self, new_words):
```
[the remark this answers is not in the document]
jinyongyoo: I think `generate_new_attacked_text` is clearer than `replace_new_words`. `replace_new_words` doesn't tell us what the propose of the sentence is. Like are replacing "new words" with something or are we replacing something with "new words"?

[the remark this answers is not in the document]
jxmorris12: ok i see what you mean

PR #131, opened 2020-06-19
Word swap using BERT

jinyongyoo:
Basic words swap using `BERTMaksedLM` from `transformers`.


textattack/transformations/word_swap_masked_lm.py
```diff
@@ -0,0 +1,209 @@
...
+            token2char_offset = tokenized_text["offset_mapping"]
+            with torch.no_grad():
+                pred_probs = self._language_model(
+                    current_ids, token_type_ids=self._segment_tensor
+                )[0][0]
+            top_probs, top_ids = torch.topk(pred_probs, self.max_candidates)
+            extra_args["top_pred_ids"] = top_ids.cpu()
+            extra_args["masked_lm_logits"] = pred_probs.cpu()
+            extra_args["current_ids"] = tokenized_text["input_ids"]
+            extra_args["token2char_offset"] = token2char_offset
```
jxmorris12, line 174: can you pass these as optional parameters? the dict is too error-prone

textattack/transformations/word_swap_masked_lm.py
```diff
@@ -0,0 +1,209 @@
...
+
+        replacement_words = []
+        for id in top_ids:
+            token = self._lm_tokenizer.convert_ids_to_tokens(id)
+            if check_if_word(token):
+                replacement_words.append(token)
+
+        return replacement_words
+
+    def _bert_attack_replacement_words(self, current_text, index, extra_args):
```
jxmorris12, line 83: pls add a docstring

textattack/transformations/word_swap_masked_lm.py
```diff
@@ -0,0 +1,209 @@
...
+        self._language_model = AutoModelWithLMHead.from_pretrained(
+            masked_language_model
+        )
+        self._language_model.to(utils.device)
+        self._segment_tensor = (
+            torch.zeros(self.max_length, dtype=torch.long).unsqueeze(0).to(utils.device)
+        )
+        self._language_model.eval()
+
+    def _bae_replacement_words(self, current_text, index):
```
jxmorris12, line 48: pls add a docstring

textattack/transformations/word_swap_masked_lm.py
```diff
@@ -0,0 +1,209 @@
...
+            index, self._lm_tokenizer.mask_token
+        )
+        ids = self._lm_tokenizer.encode(
+            masked_attacked_text.text,
+            max_length=self.max_length,
+            pad_to_max_length=True,
+        )
+
+        try:
+            masked_index = ids.index(self._lm_tokenizer.mask_token_id)
```
jxmorris12, line 60: when can this happen?

textattack/transformations/word_swap_masked_lm.py
```diff
@@ -0,0 +1,209 @@
...
+            raise ValueError(f"Unrecognized value {self.method} for `self.method`.")
+
+    def _get_transformations(self, current_text, indices_to_modify):
+        extra_args = {}
+        if self.method == "bert-attack":
+            tokenized_text = self._lm_tokenizer.encode_plus(
+                current_text.text,
+                max_length=self.max_length,
+                pad_to_max_length=True,
+                return_offsets_mapping=True,
```
jxmorris12, line 161: very cool use of the tokenizer 👍

textattack/shared/attacked_text.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, text_input, attack_attrs=None):
             )
         # Format text inputs.
         self._text_input = OrderedDict([(k, v) for k, v in self._text_input.items()])
-        self.words = words_from_text(self.text)
+        self.words, self.words2char_offset = words_from_text(self.text, words2char=True)
```
jxmorris12, line 47: If you keep this, can you rename it, and add a clarifying comment?

possible better name: `word_positions_in_text`, maybe?

textattack/transformations/word_swap_masked_lm.py
```diff
@@ -0,0 +1,209 @@
...
+
+        for i in range(len(current_ids)):
+            token_start, token_end = token2char_offset[i]
+            token = self._lm_tokenizer.convert_ids_to_tokens(current_ids[i]).replace(
+                "##", ""
+            )
+            if (
+                token_start >= word_start and token_end - 1 <= word_end
+            ) and token in target_word:
+                target_ids_pos.append(i)
```
jxmorris12, line 105: I see the problem here but I'm not convinced this is the best solution. Maybe you could encode the first part of the string and then take all the tokens after that that form a single word? (As in, take ['word'] or ['wo##' '##rd'], depending on which you find.) Maybe I'm missing why this is a technical issue.

If you can't think of a better way then maybe just put a @TODO to clean this up later :)

textattack/transformations/word_swap_bert_masked_lm.py
```diff
@@ -0,0 +1,208 @@
...
+
+class WordSwapBERTMaskedLM(WordSwap):
+    """
+    Generate potential replacements for a word using BERT-Masked LM.
+
+    Based off of two papers
+        - "BAE: BERT-based Adversarial Examples for Text Classification" (Garg et al., 2020) https://arxiv.org/abs/2004.01970
+        - "BERT-ATTACK: Adversarial Attack Against BERT Using BERT" (Li et al, 2020) https://arxiv.org/abs/2004.09984
+
+    Choose which method to use by specifying "bae" or "bert-attack" for `method` argument
```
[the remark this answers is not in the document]
jxmorris12: in the comment I mean ^

textattack/transformations/word_swap_masked_lm.py
jinyongyoo: I do have another way of doing this. Mask out the word we want to replace like in BAE function, then after tokenizing, replace Mask token id with the tokens of word we want to replace. Then, we know exactly which tokens belong to the word we want to replace. Idk if this is the best way, however, since we end up tokenizing the same text over and over again. But then again, it might not matter much since we already go through the list of ids once already.

textattack/shared/attacked_text.py
jinyongyoo: Done

textattack/transformations/word_swap_masked_lm.py
jinyongyoo: Yes I think `**kwargs` would work better

textattack/transformations/word_swap_masked_lm.py
jinyongyoo: Since default `max_length=256`, tokenizer would cut off all tokens include mask token that are past 256. If the word we want to change is past 256, then we run into this issue.

textattack/transformations/word_swap_masked_lm.py
jinyongyoo: Done

textattack/transformations/word_swap_masked_lm.py
jinyongyoo: Done

textattack/transformations/word_swap_bert_masked_lm.py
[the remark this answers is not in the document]
jinyongyoo: done

textattack/transformations/word_swap_masked_lm.py
jxmorris12: thats a very good idea :)

PR #137, opened 2020-06-20
multi-command argparse

jxmorris12:


jinyongyoo:
Should we include `TrainCommand` if it's not a full feature?

jxmorris12:
> Should we include `TrainCommand` if it's not a full feature?

incoming tonight (before we update the pip package)

PR #139, opened 2020-06-21
`train` module

jxmorris12:
Train models using TextAttack:
- support word CNN
- support LSTM
- support transformers models like `bert-base-uncased`
- automatically load `nlp` datasets
- log to tensorboard


jxmorris12:
LGTM! :)

PR #142, opened 2020-06-21
color output and show label and confidence score

jxmorris12:


jxmorris12:
outputs look kinda like this

PR #147, opened 2020-06-22
add optional dependencies

jxmorris12:


jxmorris12:
we're giving up on this for now– too much confusion for users. Someday we'll circle back around to this idea of optional dependencies.

PR #150, opened 2020-06-24
universal cache dir

jxmorris12:
- get rid of `config.yaml`
- tensorflow hub, transformers, etc. use TA_CACHE_DIR
- add `pip install "[.dev]"` option

### DIFF
--- a/textattack/commands/attack/attack_command.py
+++ b/textattack/commands/attack/attack_command.py
@@ -157,6 +157,24 @@ class AttackCommand(TextAttackCommand):
             default=float("inf"),
             help="The maximum number of model queries allowed per example attacked.",
         )
+        parser.add_argument(
+            "--model-batch-size",
+            type=int,
+            default=32,
+            help="The batch size for making calls to the model.",
+        )
+        parser.add_argument(
+            "--model-cache-size",
+            type=int,
+            default=2 ** 18,
+            help="The maximum number of items to keep in the model results cache at once.",
+        )
+        parser.add_argument(
+            "--constraint-cache-size",
+            type=int,
+            default=2 ** 18,
+            help="The maximum number of items to keep in the constraints cache at once.",
+        )
 
         attack_group = parser.add_mutually_exclusive_group(required=False)
         search_choices = ", ".join(SEARCH_METHOD_CLASS_NAMES.keys())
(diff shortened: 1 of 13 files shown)


textattack/commands/attack/attack_command.py
```diff
@@ -157,6 +157,24 @@ def str_to_int(s):
             default=float("inf"),
             help="The maximum number of model queries allowed per example attacked.",
         )
+        parser.add_argument(
```
jinyongyoo, line 160: When would the user ever need to set the batch size when attacking? We don't process attacks in batches.

textattack/shared/utils/install.py
```diff
@@ -122,28 +120,39 @@ def _post_install():
     nltk.download("stopwords")
 
 
+def set_cache_dir(cache_dir):
+    """ Sets all relevant cache directories to ``TA_CACHE_DIR``. """
+    # Tensorflow Hub cache directory
+    os.environ["TFHUB_CACHE_DIR"] = cache_dir
```
jinyongyoo, line 126: I'm not sure if this is necessarily good for the user. (1) They might not know that their cache location just got switched. (2) If they already have models and stuff cached, shouldn't we allow them to use it in the first place.

textattack/shared/utils/misc.py
```diff
@@ -86,3 +86,4 @@ def set_seed(random_seed):
     random.seed(random_seed)
     np.random.seed(random_seed)
     torch.manual_seed(random_seed)
+    torch.cuda.manual_seed(random_seed)
```
jinyongyoo, line 89: wow i didn't know we also had to set seed for CUDA. Good catch!

textattack/shared/attack.py
```diff
@@ -145,18 +148,20 @@ def _filter_transformations(
         # Populate cache with transformed_texts
         uncached_texts = []
         for transformed_text in transformed_texts:
-            if transformed_text not in self.constraints_cache:
+            if (current_text, transformed_text) not in self.constraints_cache:
```
jinyongyoo, line 151: Very good catch.

textattack/shared/utils/misc.py
jxmorris12: ikr

textattack/shared/attack.py
jxmorris12: :)

textattack/shared/utils/install.py
jxmorris12: resolved w Jeffrey: In the future we'll add multiple levels/preferences for the cache directory

PR #151, opened 2020-06-24
Goal function updates

uvafan:
- Allow for maximizing goal function, rather than satisficing
- Add "start new example" function in GoalFunction which sets num_queries to 0, sets an initial_tokenized_text
- Change CSV to log every attack result, and have column whether succeeded, failed, maximized, or skipped. Show ground truth output in CSV
- Add input reduction recipe
- Various small bug fixes that I noticed while doing the above


jxmorris12:
setting as draft while Eli adds input reduction

textattack/goal_functions/classification/input_reduction.py
```diff
@@ -0,0 +1,36 @@
+from .classification_goal_function import ClassificationGoalFunction
+
+
+class InputReduction(ClassificationGoalFunction):
+    """
+    An targeted attack on classification models which attempts to maximize the
```
jxmorris12, line 6: 1. typo: 'A targeted'


2. maybe you should mention the Feng et al paper in this docstring? If I had stumbled upon this class without knowing about that paper, I might be very confused what its use is. Just an idea.

3. nit about this docstring: might be better if this says "... attempts to maximize the score of the target label. Complete when the target label is the predicted label. " or something like that-- I had to read this a couple times to understand

textattack/goal_functions/classification/input_reduction.py
```diff
@@ -0,0 +1,36 @@
...
+            self.ground_truth_output == model_output.argmax()
+            and attacked_text.num_words <= target_num_words
+        )
+
+    def _should_skip(self, model_output, attacked_text):
+        return self.ground_truth_output != model_output.argmax()
+
+    def _get_score(self, model_output, attacked_text):
+        if self.ground_truth_output != model_output.argmax():
+            return float("-inf")
```
jxmorris12, line 25: why do this here and not in untargeted/targeted goal functions? shouldn't this be skipped anyway?

textattack/goal_functions/classification/input_reduction.py
```diff
@@ -0,0 +1,36 @@
...
+        cur_num_words = attacked_text.num_words
+        initial_num_words = self.initial_attacked_text.num_words
+        num_words_score = (initial_num_words - cur_num_words) / initial_num_words
```
jxmorris12, line 28: clamp this at 0 in case `initial_num_words < cur_num_words`, like if some dummy used this with a word insertion transformation?

textattack/goal_functions/classification/input_reduction.py
```diff
@@ -0,0 +1,36 @@
...
+        model_score = model_output[self.ground_truth_output]
+        return num_words_score + model_score / initial_num_words
```
jxmorris12, line 30: 1. can you add a comment explaining why this score is a sum of two things (model score is used as a tiebreaker)

2. this can overflow above 1, right? so maybe either [1] clamp at 1, that should be fine or [2] divide by (initial_results + 1) -- I think that would look like ```return  (initial_num_words - cur_num_words + model_score) / (initial_num_words + 1)```

textattack/goal_functions/classification/input_reduction.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def _get_displayed_output(self, raw_output):
+        return int(raw_output.argmax())
```
jxmorris12, line 33: nit: if you want, you can move this method into `ClassificationGoalFunction` and then remove it from here, untargeted, and targeted classes

textattack/loggers/csv_logger.py
```diff
@@ -20,17 +20,18 @@ def __init__(self, filename="results.csv", color_method="file"):
         self._flushed = True
 
     def log_attack_result(self, result):
-        if isinstance(result, FailedAttackResult):
-            return
         original_text, perturbed_text = result.diff_color(self.color_method)
+        result_type = result.__class__.__name__[:-12]
```
jxmorris12, line 24: haha. this is just to remove 'AttackResult' right? 

better: `result.class.__name__.replace('AttackResult', '')`

best: change the `GoalFunctionResultStatus` enum to map to a string (like `SUCCEEDED = 'succeeded'` instead of `SUCCEEDED = 0`) and then use `result.perturbed_result.status` (or something like that)

textattack/search_methods/greedy_word_swap_wir.py
```diff
@@ -151,7 +144,7 @@ def check_transformation_compatibility(self, transformation):
         """
             Since it ranks words by their importance, GreedyWordSwapWIR is limited to word swaps transformations.
```
jxmorris12, line 145: change to `"... is limited to word swap and deletion transformations."`

textattack/shared/attack.py
```diff
@@ -176,17 +178,18 @@ def attack_one(self, initial_result):
             initial_result: The initial ``GoalFunctionResult`` from which to perturb.
 
         Returns:
-            Either a ``SuccessfulAttackResult`` or ``FailedAttackResult``.
+            Either a ``SuccessfulAttackResult``, ``FailedAttackResult``,
```
jxmorris12, line 181: nit: can you remove `Either` since it's a list of 3 now lol

textattack/shared/attacked_text.py
```diff
@@ -376,6 +376,11 @@ def text(self):
         """
         return "\n".join(self._text_input.values())
 
+    @property
+    def num_words(self):
+        """ Returns the number of words in the sequence. """
+        return len(self.words)
+
```
jxmorris12, line 383: I think this is worth being its own method - but there's probably a lot of uses of `len(at.words)` still scattered about the codebase, right?

if so, can you put a task on asana and/or github issue to clean up the code and replace them all with `at.num_words`?

textattack/goal_functions/goal_function.py
```diff
@@ -42,38 +49,40 @@ def __init__(
...
-        self.num_queries = 0
         self.query_budget = query_budget
         self.model_batch_size = model_batch_size
         if self.use_cache:
             self._call_model_cache = lru.LRU(model_cache_size)
         else:
             self._call_model_cache = None
 
-    def should_skip(self, attacked_text, ground_truth_output):
+    def init_attack_example(self, attacked_text, ground_truth_output):
```
jxmorris12, line 59: [discussion about this on Slack]

textattack/goal_functions/classification/input_reduction.py
uvafan: Well the goal of input reduction is to make sure the class score isn't changed. It won't be skipped if the class score is changed through the process of removing words. This isn't in untargeted/targeted goal functions because there isn't the same sort of "maximize X conditioned on Y" sort of property that input reduction has ("maximize words removed conditioned on the predicted class not changing")

textattack/goal_functions/classification/input_reduction.py
uvafan: I just forgot to change this from the TargetedClassification class LOL

textattack/goal_functions/goal_function.py
uvafan: Added an Asana task

PR #159, opened 2020-06-28
Fix genetic

jinyongyoo:


textattack/shared/attack.py
```diff
@@ -77,6 +77,9 @@ def __init__(
         self.search_method.get_transformations = self.get_transformations
         self.search_method.get_goal_results = self.goal_function.get_results
 
+        if isinstance(self.search_method, textattack.search_methods.GeneticAlgorithm):
+            self.search_method.filter_transformations = self.filter_transformations
```
jxmorris12, line 81: maybe we could just pass `filter_transformations` as an argument every time? I don't see a problem w that

textattack/constraints/semantics/sentence_encoders/sentence_encoder.py
```diff
@@ -153,7 +153,7 @@ def _score_list(self, starting_text, transformed_texts):
                 # If the embedding is not yet a tensor, make it one.
                 starting_embedding = torch.tensor(embeddings[0]).to(utils.device)
 
-            if isinstance(embeddings, list):
+            if isinstance(embeddings[1], torch.Tensor):
```
jxmorris12, line 156: why?

textattack/search_methods/genetic_algorithm.py
```diff
@@ -19,184 +19,219 @@ class GeneticAlgorithm(SearchMethod):
...
     """
 
     def __init__(
-        self, pop_size=20, max_iters=50, temp=0.3, give_up_if_no_improvement=False
+        self,
+        pop_size=20,
+        max_iters=50,
+        temp=0.3,
+        give_up_if_no_improvement=False,
+        max_crossover_retries=20,
```
jxmorris12, line 37: this default seems fine to me, but can you set it to 0 in the alzantot recipe?

textattack/search_methods/genetic_algorithm.py
```diff
@@ -19,184 +19,219 @@ class GeneticAlgorithm(SearchMethod):
...
+                break
+
+            num_tries += 1
+
+        if not passed_constraints:
+            new_text = (
+                pop_member1.attacked_text
+                if np.random.uniform() < 0.5
+                else pop_member2.attacked_text
+            )
```
jxmorris12, line 150: I think this makes sense, but can you add a comment explaining what this is doing?

textattack/search_methods/genetic_algorithm.py
```diff
@@ -19,184 +19,219 @@ class GeneticAlgorithm(SearchMethod):
...
+                    initial_result,
+                )
+
+                self._perturb(child, initial_result)
+                if child.result is None:
+                    # If child.result is not computed for any reason, we compute it here.
+                    result, self._search_over = self.get_goal_results(
+                        [child.attacked_text], initial_result.output
+                    )
+                    child.result = result[0]
```
jxmorris12, line 224: when can this happen..?

textattack/shared/attack.py
jinyongyoo: I guess we could. I was wondering if that would leave people confused as to whether they would need to `filter_transformations` after they get transformations from `get_transformations`

textattack/constraints/semantics/sentence_encoders/sentence_encoder.py
jinyongyoo: I don't know why we do `instance(embeddings, list)` check. But to do `torch.stack`, the elements themselves need to be a Tensor. I'm just here assuming that checking `embeddings[1]` is good enough since it's batch output. If you run it with `bert` sentence encoder, I'm see that `embedding` is a list of `np.array`. In that case, we should just do `torch.tesnor`.

jxmorris12: ok I think it's bc the old sentence encoder outputted a list

textattack/search_methods/genetic_algorithm.py
```diff
@@ -19,184 +19,219 @@ class GeneticAlgorithm(SearchMethod):
...
+
+        return PopulationMember(new_text, num_neighbors_list)
+
+    def _initialize_population(self, initial_result):
         """
-        Generates this neighbors_len list
+        Initialize a population of texts each with one word replaced
         Args:
-            attacked_text: The original text
+            initial_result (GaolFunctionResult): The result to instantiate the population with
```
jxmorris12, line 158: typo (GoalFunctionResult)

PR #173, opened 2020-06-30
improve word delimiting logic

jxmorris12:
- support apostrophes contained within words: `doesn't` is now 1 word, not 2
- support hyphens contained within words: `well-placed` is now 1 word, not 2

minor updates:
- update to transformers 3.x syntax
- fix value error with `PWWS` recipe
- fix type errors with `kuleshov` recipe 
  - & add test so it doesn't happen again
- github runner updates
  - separate runners for test, format, and docs
  - fix docs syntax errors :)

depends on Eli's PR #151


jxmorris12:
just reviewed my own PR and going to merge now! :)

PR #174, opened 2020-06-30
PSO model is ready

jind11:
this branch is for the PSO model, now it is ready to be reviewed and merged


textattack/attack_recipes/PSO_zang_2020.py
```diff
@@ -0,0 +1,45 @@
...
+    # input_column_modification = InputColumnModification(
+    #     ["premise", "hypothesis"], {"premise"}
+    # )
+    # constraints.append(input_column_modification)
+    #
+    # Goal is untargeted classification
+    #
+    goal_function = UntargetedClassification(model)
+    #
+    # Perform word substitution with a genetic algorithm.
```
jxmorris12, line 41: I think this recipe file should quote the paper:

> We propose a novel word substitution-based textual attack model, which reforms both the aforementioned two steps. In the first step, we adopt a sememe-based word substitution strategy, which can generate more candidate adversarial examples with better semantic preservation. In the second step, we utilize particle swarm optimization (Eberhart and Kennedy, 1995) as the adversarial example searching algorithm"

textattack/search_methods/PSO_algorithm.py
```diff
@@ -0,0 +1,266 @@
...
+from copy import deepcopy
+
+import numpy as np
+
+from textattack.search_methods import SearchMethod
+
+
+class PSOAlgorithm(SearchMethod):
+    """
+    Attacks a model with word substiutitions using a PSO algorithm.
```
jxmorris12, line 17: can you say "using a Particle Swarm Optimization algorithm" here, since some people might stumble upon this file and not know what PSO stands for.

textattack/search_methods/PSO_algorithm.py
```diff
@@ -0,0 +1,266 @@
...
+
+    Args:
+        pop_size (:obj:`int`, optional): The population size. Defauls to 20.
+        max_iters (:obj:`int`, optional): The maximum number of iterations to use. Defaults to 50.
```
jxmorris12, line 21: these comments should say "Defaults to 60", and "Defaults to 20"

textattack/search_methods/PSO_algorithm.py
```diff
@@ -0,0 +1,266 @@
...
+                indices_to_replace.append(i)
+                words_to_replace.append(x2_words[i])
+        new_text = x1.replace_words_at_indices(indices_to_replace, words_to_replace)
+        return new_text
+
+    def count_change_ratio(self, x1, x2, x_len):
+        change_ratio = float(np.sum(x1.words != x2.words)) / float(x_len)
+        return change_ratio
+
+    def sigmod(self, n):
```
jxmorris12, line 152: typo: `sigmod -> sigmoid`

textattack/search_methods/PSO_algorithm.py
```diff
@@ -0,0 +1,266 @@
...
+        all_elite_score = pop_scores[top_attack]
+        if pop_results[top_attack].succeeded:
+            return pop_results[top_attack]
+
+        # set up hyper-parameters
+        Omega_1 = 0.8
+        Omega_2 = 0.2
+        C1_origin = 0.8
+        C2_origin = 0.2
+        V = [np.random.uniform(-3, 3) for _ in range(self.pop_size)]
```
jxmorris12, line 187: should this `3` value be a property of the class? (`self.V_max` or something like that)

nit: `np.random.uniform(-3, 3, self.pop_size)` should work here, I think

textattack/search_methods/PSO_algorithm.py
```diff
@@ -0,0 +1,266 @@
...
+        all_elite = pop[top_attack]
+        all_elite_score = pop_scores[top_attack]
+        if pop_results[top_attack].succeeded:
+            return pop_results[top_attack]
+
+        # set up hyper-parameters
+        Omega_1 = 0.8
+        Omega_2 = 0.2
+        C1_origin = 0.8
+        C2_origin = 0.2
```
jxmorris12, line 186: I think these should all be properties of the class, initialized in `__init__`

```
self.Omega_1 = ...
```

because someone might want to try running this PSO algorithm with different hyperparameters!

textattack/transformations/word_swap_hownet.py
```diff
@@ -0,0 +1,103 @@
...
+            ]
+        except KeyError:
+            # This word is not in our synonym bank, so return an empty list.
+            return []
+
+    def _get_transformations(self, current_text, indices_to_modify):
+        words = current_text.words
+        words_str = " ".join(words)
+        word_list, pos_list = zip_flair_result(
+            self._flair_pos_tagger.predict(words_str)[0]
```
jxmorris12, line 56: `words_str` seems to just be `current_text.text` with the punctuation removed

why not `zip_flair_result(self._flair_pos_tagger.predict(current_text.text)[0])`?

textattack/transformations/word_swap_hownet.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+        assert len(words) == len(word_list)
```
jxmorris12, line 58: can you add a helpful error message to this `assert` please, something like 

```
assert len(words) == len(word_list), "Part-of-speech tagger returned incorrect number of tags"
```

textattack/transformations/word_swap_hownet.py
```diff
@@ -0,0 +1,103 @@
...
+    """ Transforms an input by replacing its words with synonyms in the stored synonyms bank
+    generated by the OpenHowNet. """
+
+    PATH = "hownet"
+
+    def __init__(self, max_candidates=-1, **kwargs):
+        super().__init__(**kwargs)
+        self.max_candidates = max_candidates
+
+        # Download embeddings if they're not cached.
```
jxmorris12, line 20: typo: embeddings -> candidates (or something)

textattack/search_methods/PSO_algorithm.py
```diff
@@ -0,0 +1,266 @@
...
+                assert all(
+                    [
+                        w1 == w2
+                        for w1, w2 in zip(attacked_text.words, transformed_text.words)
+                    ]
+                )
+        neighbors_list = [np.array(x) for x in neighbors_list]
+        return neighbors_list
+
+    def equal(self, a, b):
```
jxmorris12, line 131: **private methods**
can you make the methods that will never be called from outside of this class private (prefix with `_`?) so:
- `equal -> _equal`
- `turn -> _turn`
- `sigmoid -> _sigmoid`
- `generate_population -> _generate_population`
- `gen_h_score -> _gen_h_score`

textattack/search_methods/PSO_algorithm.py
```diff
@@ -0,0 +1,266 @@
...
+        s = np.sum(tn)
+        if s == 0:
+            for i in range(len(tn)):
+                tn[i] = 1
+            return [t / len(tn) for t in tn]
+        new_n = [t / s for t in tn]
+
+        return new_n
+
+    # for un-targeted attacking
```
jxmorris12, line 75: is this just for un-targeted attacking? won't "goal results" work for any goal?

textattack/search_methods/PSO_algorithm.py
```diff
@@ -0,0 +1,266 @@
...
+                else:
+                    new_pop.append(x)
+            pop = new_pop
+
+            # check if there is any successful attack in the current population
+            pop_results, self.search_over = self.get_goal_results(
+                pop, self.correct_output
+            )
+            if self.search_over:
+                return max(pop_results, key=lambda x: x.score)
```
jxmorris12, line 251: good way to do it!

textattack/transformations/word_swap_hownet.py
```diff
@@ -0,0 +1,103 @@
...
+    elif reference_word.isupper() and len(reference_word) > 1:
+        return word.upper()
+    elif reference_word[0].isupper() and reference_word[1:].islower():
+        return word.capitalize()
+    else:
+        # if other, just do not alter the word's case
+        return word
+
+
+def zip_flair_result(pred):
```
jxmorris12, line 92: can you add a comment explaining what this function does?

textattack/search_methods/PSO_algorithm.py
jind11: it can be used for targeted attacking, but we should choose either the untargeted attacking or targeted one for this recipe, and isn't the untargeted attack the default one for demo?

textattack/transformations/word_swap_hownet.py
jind11: since I need to get the POS tag for each word in the "words" list, and from the text to words list, many non-alphabet things have been removed, I do not know map from the POS results of text to each word in the words list.

jxmorris12:
```
for attacking 100 examples of SST-2:
Number of successful attacks: | 90   |
| Number of failed attacks:   | 2    |
| Number of skipped attacks:  | 8    |
| Original accuracy:      | 92.0%  |
| Accuracy under attack:    | 2.0%  |
| Attack success rate:     | 97.83% |
| Average perturbed word %:   | 13.25% |
| Average num. words per input: | 17.14  |
| Avg num queries:       | 9201.83
```

PR #182, opened 2020-07-03
IGA 

sherlockyyc:
Adds the Improved Genetic Algorithm attack (https://arxiv.org/abs/1909.06723)

- Adding the IGA as a search method.
- Creating an "Attack Recipe" for constructing the full attack from TextAttack components. 
- Editing the attack_args_helper file (specifically, the RECIPE_NAMES variable) to add the ability to run the IGA from the command-line.


jxmorris12:
@sherlockyyc this looks great!! will review and let you know if there are any problems within a few days.

textattack/search_methods/improved_genetic_algorithm.py
```diff
@@ -0,0 +1,257 @@
...
+                if np.random.uniform() < 0.5
+                else pop_member2.attacked_text
+            )
+
+        new_results, self._search_over = self.get_goal_results([new_text])
+
+        return PopulationMember(new_text, new_replaced_times, new_results[0])
+
+    def _perform_search(self, initial_result):
+        self.original_attacked_text = initial_result.attacked_text
```
jinyongyoo, line 186: Two things. There should be `self._search_over = False` before every search just so that we don't happen to prematurely end search if `self._search_over=True` from previous search. Also, do you mind removing `self.original_attacked_text` and switch to passing it as explicit function argument? This way, we minimize the "stateful-ness" of the search method.

textattack/search_methods/improved_genetic_algorithm.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def _mutate(self, pop_member):
+        """
+        Replaces a word in pop_member that has not been modified. 
+        Args:
+            pop_member: The population member being perturbed.
+        """
+        x_len = pop_member.replaced_times.shape[0]
+        iterations = 0
+        while iterations < x_len and not self._search_over:
```
jinyongyoo, line 91: One question. So I see that we try sampling `n` number of times where `n` is number of words. What happens if we for some reason end up unlucky and sample the same few indices that fail to produce transformations that improve the score? Should there be something that says "hey, don't sample this indices because it's bad". In the original GA method, this is done by explicitly defining the discrete probability distribution.

textattack/attack_recipes/iga_wang_2019.py
```diff
@@ -0,0 +1,55 @@
...
+from textattack.search_methods import ImprovedGeneticAlgorithm
+from textattack.shared.attack import Attack
+from textattack.transformations import WordSwapEmbedding
+
+
+def IGAWang2019(model):
+    """
+        Xiaosen Wang, Hao Jin, Kun He (2019). 
+        
+        Natural Language Adversarial Attack and Defense inWord Level.
```
jinyongyoo, line 20: Space for `inWord`

textattack/attack_recipes/iga_wang_2019.py
```diff
@@ -0,0 +1,55 @@
...
+        Natural Language Adversarial Attack and Defense inWord Level. 
+        
+        http://arxiv.org/abs/1909.06723 
+    """
+    #
+    # Swap words with their embedding nearest-neighbors.
+    #
+    # Embedding: Counter-fitted Paragram Embeddings.
+    #
+    # "[We] fix the hyperparameter values to S = 60, δ = 0.5, λ = 5, and N is unrestricted (50)."
```
jinyongyoo, line 29: Can you move this comment to be above where you define `search_method`. Right now it looks like it's about `WordSwapEmbedding`

sherlockyyc: Thanks for your comments, I will be revising the code in the next few days.

sherlockyyc:
> Overall, this looks good! Thank you for the contribution. I wonder if it is possible to make your version of GA appear similar to the original version **as in code-wise**. For instance, we can make them extend the same abstract GA class and implement abstract methods and/or share `PopulationMember` class.
> 
> This isn't to say you need to make your version look exactly like ours. We can also make adjustments to the original GA. But I think it'll be helpful to make the similarity between the two as clear as possible, while also highlighting their differences.

According to your suggestion, I recommitted the code. GA and IGA differ in only a few details, so I just added a menber variable `improved_genetic_algorithm (bool)`  in class `GeneticAlgorithm` to determine whether to use IGA, and explicitly highlighted their different treatments in several details through If-else code. I don't know if this is the right way to implement IGA as a new recipe. If you think it is a better solution to let GA and IGA extend the same abstract class, then I can also modify the code again. Thank you for your review.

sherlockyyc:
@jinyongyoo According to your suggestion, I recommitted the code. GA and IGA differ in only a few details, so I just added a menber variable improved_genetic_algorithm (bool) in class GeneticAlgorithm to determine whether to use IGA, and explicitly highlighted their different treatments in several details through If-else code. I don't know if this is the right way to implement IGA as a new recipe. If you think it is a better solution to let GA and IGA extend the same abstract class, then I can also modify the code again. Thank you for your review.

jinyongyoo:
@sherlockyyc thank you for the update! I'll take a look at it in the upcoming days and let you know!

jxmorris12:
@sherlockyyc btw, sorry, the tests were failing before. Some packages got updated that broke our master branch. It should all be fixed now.

PR #189, opened 2020-07-06
minor bug fixes

jxmorris12:
- fix flair version bug and fix version; 
- fix isort version bug and fix verison;
- fix alzantot abstract constraint class bug

### DIFF
--- a/textattack/constraints/grammaticality/language_models/google_language_model/google_language_model.py
+++ b/textattack/constraints/grammaticality/language_models/google_language_model/google_language_model.py
@@ -51,6 +51,9 @@ class GoogleLanguageModel(Constraint):
 
         def get_probs(current_text, transformed_texts):
             word_swap_index = current_text.first_word_diff_index(transformed_texts[0])
+            if word_swap_index is None:
+                return []
+
             prefix = current_text.words[word_swap_index - 1]
             swapped_words = np.array(
                 [t.words[word_swap_index] for t in transformed_texts]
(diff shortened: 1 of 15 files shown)


textattack/constraints/grammaticality/language_models/google_language_model/google_language_model.py
```diff
@@ -75,6 +78,9 @@ def get_probs(current_text, transformed_texts):
             probs_of_swaps_at_index = list(
                 zip(item_indices, get_probs(current_text, this_transformed_texts))
             )
+            # if len(probs_of_swaps_at_index) == 0:
+            #     probs.extend(0)
+            #     continue
```
heytitle, line 83: if we don't need this block here, maybe we can remove it?

textattack/constraints/grammaticality/language_models/google_language_model/google_language_model.py
```diff
@@ -75,6 +78,9 @@ def get_probs(current_text, transformed_texts):
             probs_of_swaps_at_index = list(
                 zip(item_indices, get_probs(current_text, this_transformed_texts))
             )
+            # if len(probs_of_swaps_at_index) == 0:
+            #     probs.extend(0)
```
jinyongyoo, line 82: Should this be commented?

textattack/constraints/grammaticality/language_models/google_language_model/google_language_model.py
jxmorris12: thanks! will delete

textattack/constraints/grammaticality/language_models/google_language_model/google_language_model.py
jxmorris12: thanks! will delete

PR #194, opened 2020-07-07
update github runner

jxmorris12:


jxmorris12:
in the end, it wasn't poetry, but at least it fixed our tests

PR #196, opened 2020-07-08
AG News

jxmorris12:


textattack/commands/train_model/run_training.py
```diff
@@ -171,6 +172,12 @@ def is_writable_type(obj):
 
     args_dict = {k: v for k, v in vars(args).items() if is_writable_type(v)}
 
+    # Save original args to file
+    args_save_path = os.path.join(args.output_dir, "train_args.json")
+    with open(args_save_path, "w", encoding="utf-8") as f:
+        f.write(json.dumps(args_dict, indent=2) + "\n")
+    logger.info(f"Wrote original training args to {args_save_path}.")
+
```
jind11, line 180: why should we save the args twice?

PR #204, opened 2020-07-09
augmentation updates, adversarial training

jakegrigsby:
Based on `textattack train --attack *recipe*`, TextAttack will run an attack against the current model at the beginning of each new training epoch, and use the resulting adversarial examples as training data for that epoch. This should improve robustness. Benchmarks coming soon.

Also allows for a combination of data augmentation and adversarial training (and fixes augmentation of eval set bug)

### DIFF
--- a/textattack/augmentation/augmenter.py
+++ b/textattack/augmentation/augmenter.py
@@ -75,7 +75,7 @@ class Augmenter:
         attacked_text = AttackedText(text)
         original_text = attacked_text
         all_transformed_texts = set()
-        num_words_to_swap = int(self.pct_words_to_swap * len(attacked_text.words))
+        num_words_to_swap = max(int(self.pct_words_to_swap * len(attacked_text.words)), 1)
         for _ in range(self.transformations_per_example):
             index_order = list(range(len(attacked_text.words)))
             random.shuffle(index_order)
--- a/textattack/augmentation/augmenter.py
+++ b/textattack/augmentation/augmenter.py
@@ -137,3 +137,6 @@ class Augmenter:
             all_text_list.extend([text] + augmented_texts)
             all_id_list.extend([_id] * (1 + len(augmented_texts)))
         return all_text_list, all_id_list
+
+    def __repr__(self):
+        return type(self).__name__
--- a/textattack/augmentation/augmenter.py
+++ b/textattack/augmentation/augmenter.py
@@ -3,7 +3,7 @@ import random
 import tqdm
 
 from textattack.constraints import PreTransformationConstraint
-from textattack.shared import AttackedText
+from textattack.shared import AttackedText, utils
 
 
 class Augmenter:
@@ -75,7 +75,9 @@ class Augmenter:
         attacked_text = AttackedText(text)
         original_text = attacked_text
         all_transformed_texts = set()
-        num_words_to_swap = max(int(self.pct_words_to_swap * len(attacked_text.words)), 1)
+        num_words_to_swap = max(
+            int(self.pct_words_to_swap * len(attacked_text.words)), 1
+        )
         for _ in range(self.transformations_per_example):
             index_order = list(range(len(attacked_text.words)))
             random.shuffle(index_order)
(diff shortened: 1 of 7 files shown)


jakegrigsby:
adversarial training still needs benchmarking, but this has to be merged to fix augmenation-related bugs...

textattack/augmentation/augmenter.py
```diff
@@ -137,3 +137,6 @@ def augment_text_with_ids(self, text_list, id_list, show_progress=True):
             all_text_list.extend([text] + augmented_texts)
             all_id_list.extend([_id] * (1 + len(augmented_texts)))
         return all_text_list, all_id_list
+
+    def __repr__(self):
+        return type(self).__name__
```
jxmorris12, line 142: can you copy the `Attack` repr to print `self.transformation` and `self.constraints` please

textattack/commands/train_model/run_training.py
```diff
@@ -16,24 +17,151 @@
...
 device = textattack.shared.utils.device
 logger = textattack.shared.logger
 
+def save_args(args, save_path):
+    # Save args to file
+    final_args_dict = {k: v for k, v in vars(args).items() if is_writable_type(v)}
+    with open(save_path, "w", encoding="utf-8") as f:
+        f.write(json.dumps(final_args_dict, indent=2) + "\n")
+
+def get_sample_count(*lsts):
```
jxmorris12, line 33: please add a docstring

textattack/commands/train_model/run_training.py
```diff
@@ -16,24 +17,151 @@
...
+    return sample_count
+
+def random_shuffle(*lsts):
+    permutation = np.random.permutation(len(lsts[0]))
+    shuffled = []
+    for lst in lsts:
+        shuffled.append((np.array(lst)[permutation]).tolist())
+    return tuple(shuffled)
+
+def train_val_split(*arrays, split_val=.2):
```
jxmorris12, line 47: please add a docstring

textattack/commands/train_model/run_training.py
```diff
@@ -16,24 +17,151 @@
...
+    arrays = random_shuffle(*arrays)
+    split_idx = math.floor(sample_count * split_val)
+    train_set = [array[split_idx:] for array in arrays]
+    val_set = [array[:split_idx] for array in arrays]
+    if len(train_set) == 1 and len(val_set) == 1:
+        train_set = train_set[0]
+        val_set = val_set[0]
+    return train_set, val_set
+
+def filter_labels(text, labels, allowed_labels):
```
jxmorris12, line 60: please add a docstring [sorry]

example of usage -- SNLI has -1 when the label is unclear. normal label set is 0,1,2. so users can specify --allowed-labels 0 1 2 and automatically skip train/eval samples with a -1 label

textattack/commands/train_model/run_training.py
```diff
@@ -16,24 +17,151 @@
...
+    text_ids = batch_encode(tokenizer, text)
+    input_ids = np.array(text_ids)
+    labels = np.array(labels)
+    data = list((ids, label) for ids, label in zip(input_ids, labels))
+    sampler = RandomSampler(data)
+    dataloader = DataLoader(data, sampler=sampler, batch_size=batch_size)
+    return dataloader
+
+def data_augmentation(text, labels, augmenter):
+    aug_text = augmenter.augment_many(text)
```
jxmorris12, line 149: please add a docstring

textattack/commands/train_model/run_training.py
```diff
@@ -16,24 +17,151 @@
...
+    flat_aug_text = []
+    flat_aug_labels = []
+    for i, examples in enumerate(aug_text):
+        for aug_ver in examples:
+            flat_aug_text.append(aug_ver)
+            flat_aug_labels.append(labels[i])
+    return flat_aug_text, flat_aug_labels
+
+def attack_model(model, attack_t, dataset):
+    attack = attack_t(model)
```
jxmorris12, line 160: please add a docstring [what's attack_t?]

textattack/commands/train_model/run_training.py
```diff
@@ -23,24 +25,253 @@
...
+    :param dataset: iterable of (text, label) pairs.
+
+    :return: list of adversarial examples.
+    """
+    attack = attackCls(model)
+    adv_train_text = []
+    for adv_ex in tqdm.tqdm(
+        attack.attack_dataset(dataset), desc="Attack", total=len(dataset)
+    ):
+        adv_train_text.append(adv_ex.perturbed_text())
```
uvafan, line 265: shouldn't this check that adv_ex is a `SuccessfulAttackResult` or `MaximizedAttackResult`?

jakegrigsby: Ah, maybe. if `FailedAttackResult`, and add `original_text()`, else add `perturbed_text()`?

uvafan: Yeah that sounds right. Could maybe have boolean options for:
- Whether to include or ignore Failed/skipped attack results in the adv training epochs
- Whether to use `original_text()` or as far as the attack got (`perturbed_text()`) for `FailedAttackResult`s

jakegrigsby: If this was image adversarial training, we'd include the perturbed version of failed attacks, because they are still the 'worst case' input we can come up with under our own constraints. I don't think you'd want to revert to the clean/best-case. Only problem would be if the perturbed text of a failed result was so distorted by repeated attempts that it becomes a bad sample, but from what I can tell the FailedAttackResult contains the most successful attack and wouldn't run into this...

jakegrigsby:
adversarial training will keep getting tweaked, but I'm merging to fix all the bugs that were discovered along the way.

PR #217, opened 2020-07-14
--save-last option for training

jakegrigsby:
quick change to the training code I was using to get the adversarial training results for the paper. `textattack train --save-last` saves the model weights to disk after the last epoch, instead of only preserving the highest eval acc. When adversarial training you're often more interested in the robustness of the model after a specific number of epochs, not in getting the highest possible eval acc
.

### DIFF
--- a/textattack/commands/train_model/run_training.py
+++ b/textattack/commands/train_model/run_training.py
@@ -552,6 +552,9 @@ def train_model(args):
         f"Eval of saved model {'pearson correlation' if args.do_regression else 'accuracy'}: {eval_score*100}%"
     )
 
+    if args.save_last:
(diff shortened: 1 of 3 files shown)


jakegrigsby:
Yeah I agree. That may have to wait until we move away from the command line, because turning bool flags on and off based on the other ones is kinda messy. 

I want to move all of the argparsing logic up a level to `TrainModelCommand.run`, and have the `train_model(args)` signature changed to be kwarg based. That way it's much easier to use it in new scripts without having to build a fake argparser. I started making those changes earlier this week but ran into some issues with the logging system (which relies on adding new variables to `args` and then dumping the whole thing to a json). Hopefully will get back to it soon.

PR #219, opened 2020-07-19
Updated csv&txt logger naming (default & specific)

Hanyu-Liu-123:
Improve naming for CSV logger and TXT logger from `attack-[timestamp]` to `[model]-[recipe]-[timestamp]`

Enable the option to choose specific output filename by using command  `--csv-filename   [desired filename]` or `--txt-filename [desired filename]`

Example: `textattack attack --recipe textfooler --model bert-base-uncased-mr --num-examples 100 --enable-csv -l` gives filenames __CSV_AttackResult_bert-base-uncased-mr_textfooler_2020-7-18-18-2.csv__ and  __TXT_AttackResult_bert-base-uncased-mr_textfooler_2020-7-18-18-2.txt__ respectively, when called at 18:02 on 2020-7-18


textattack/commands/attack/attack_args_helpers.py
```diff
@@ -414,7 +414,8 @@ def parse_dataset_from_args(args):
 def parse_logger_from_args(args):
     # Create logger
     attack_log_manager = textattack.loggers.AttackLogManager()
-    out_time = int(time.time() * 1000)  # Output file
+    # Get current time for file naming
+    year, month, day, hour, minute = map(int, time.strftime("%Y %m %d %H %M").split())
```
uvafan, line 418: Why do you split then convert to ints when we convert it back to a string later. Just format the string how you want it here?

textattack/commands/attack/attack_command.py
```diff
@@ -81,6 +81,13 @@ def register_subcommand(main_parser: ArgumentParser):
             help="A directory to output results to.",
         )
 
+        parser.add_argument(
+            "--txt-filename",
```
uvafan, line 85: Maybe instead of adding txt/csv filename arguments separate from out-dir, we can do something like turn out-dir an --out-file argument that performs the functionality of out-dir if it ends in /, but otherwise is treated as a filename rather than directory. And maybe we can just change the extension based on txt vs. csv?

textattack/commands/attack/attack_args_helpers.py
```diff
@@ -427,13 +428,34 @@ def parse_logger_from_args(args):
...
     # if "--log-to-file" specified in terminal command, then save it to a txt file
     if args.log_to_file:
-        # Output file.
-        outfile_name = "attack-{}.txt".format(out_time)
+        # if "--txt-filename" specified, use that as filename
+        if args.txt_filename:
+            outfile_name = "{}.txt".format(args.txt_filename)
+        else:
+            if args.recipe:
+                outfile_name = "TXT_AttackResult_{}_{}_{}-{}-{}-{}-{}.txt".format(
```
uvafan, line 436: I don't think we need the "TXT_AttackResult_" part. Same goes for "CSV_AttackResult_" below. Also use [f-strings](https://www.geeksforgeeks.org/formatted-string-literals-f-strings-python/)?

textattack/commands/attack/attack_command.py
Hanyu-Liu-123: Hey Eli, I've merged `--out-dir`, `--txt-filename`, and `--csv-filename` into `--out-file-txt` and `--out-file-csv`

when `--out-file-txt` and `--out-file-csv` arguments don't end with "/", they are used to choose specific filename

when they do end with "/", they work as `--out-dir`

is that what you mean?  please let me know!

textattack/commands/attack/attack_args_helpers.py
```diff
@@ -414,28 +414,54 @@ def parse_dataset_from_args(args):
...
-        current_dir = os.path.dirname(os.path.realpath(__file__))
-        outputs_dir = os.path.join(
-            current_dir, os.pardir, os.pardir, os.pardir, "outputs", "attacks"
-        )
-        if not os.path.exists(outputs_dir):
-            os.makedirs(outputs_dir)
-        args.out_dir = os.path.normpath(outputs_dir)
+
+    # Get current time for file naming
+    year, month, day, hour, minute = map(str, time.strftime("%Y %m %d %H %M").split())
```
uvafan, line 419: these two lines can be condensed into: `timestamp = time.strftime("%Y-%m-%d-%H-%M")`

textattack/commands/attack/attack_command.py
uvafan: (Replied in the general comment in my below review)

textattack/commands/attack/attack_args_helpers.py
```diff
@@ -414,28 +414,54 @@ def parse_dataset_from_args(args):
...
+        os.makedirs(outputs_dir)
+    out_dir_txt = out_dir_csv = os.path.normpath(outputs_dir)
+
+    # Get default txt and csv file names
+    if args.recipe:
+        filename_txt = f"{args.model}_{args.recipe}_{timestamp}.txt"
+        filename_csv = f"{args.model}_{args.recipe}_{timestamp}.csv"
+    else:
+        filename_txt = f"{args.model}-{timestamp}.txt"
+        filename_csv = f"{args.model}-{timestamp}.csv"
```
heytitle, line 437: What do you think if we make the two namings here consistent (choosing either between dash `-` and underscore `_`)?

Hanyu-Liu-123: Thank you! Didn't notice that earlier...

uvafan:
LGTM, but haven't tested. Go ahead and merge if you have :)

PR #221, opened 2020-07-20
Clean up greedy-wir

uvafan:
Just a little cleanup:
- Moving the calculation of `index_order` to its own function
- Changing around a few variable names
- Removing some extra blank lines.

### DIFF
--- a/textattack/attack_recipes/pwws_ren_2019.py
+++ b/textattack/attack_recipes/pwws_ren_2019.py
@@ -25,5 +25,5 @@ def PWWSRen2019(model):
     constraints = [RepeatModification(), StopwordModification()]
     goal_function = UntargetedClassification(model)
     # search over words based on a combination of their saliency score, and how efficient the WordSwap transform is
-    search_method = GreedyWordSwapWIR("pwws")
+    search_method = GreedyWordSwapWIR("saliency")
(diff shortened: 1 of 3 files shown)


textattack/attack_recipes/pwws_ren_2019.py
```diff
@@ -25,5 +25,5 @@ def PWWSRen2019(model):
     constraints = [RepeatModification(), StopwordModification()]
     goal_function = UntargetedClassification(model)
     # search over words based on a combination of their saliency score, and how efficient the WordSwap transform is
-    search_method = GreedyWordSwapWIR("pwws")
+    search_method = GreedyWordSwapWIR("saliency")
```
jinyongyoo, line 28: I'm not sure if I'm fully onboard with this change. I can see that `pwws` is ambiguous, but "saliency" only describe just half of the actual method. The "saliency score" that the paper talks about really is just the score difference occurred by replacing words with `unk`. And as you say in the comment, the weighting of how efficient the word swaps is the other part. Maybe like `weighted-saliency` is a bit clearer?

uvafan: Good call, I will change to `weighted-saliency`

PR #223, opened 2020-07-22
add cache clearing

jinyongyoo:
Clear caches of various components after a single attack to reduce memory usage.


jinyongyoo:
No but it should not have any effect since we wouldn't use caches from attack sample 1 when attacking sample 2.

PR #226, opened 2020-07-24
Augment Interactively

Hanyu-Liu-123:
Run augmentations interactively just like running attacks

### DIFF
--- a/textattack/commands/augment.py
+++ b/textattack/commands/augment.py
@@ -28,19 +28,78 @@ class AugmentCommand(TextAttackCommand):
 
         Preserves all columns except for the input (augmneted) column.
         """
-        if not args.interactive:
+        if args.interactive:
+
+            print("\nRunning in interactive mode...\n")
+            augmenter = eval(AUGMENTATION_RECIPE_NAMES[args.recipe])(
+                pct_words_to_swap=args.pct_words_to_swap,
+                transformations_per_example=args.transformations_per_example,
+            )
+            print("--------------------------------------------------------")
+
+            while True:
+                print(
+                    '\nEnter a sentence to augment, "q" to quit, "c" to change arguments:\n'
+                )
+                text = input()
+
+                if text == "q":
+                    break
+
+                elif text == "c":
+                    print("\nChanging augmenter arguments...\n")
+                    recipe = input(
+                        "\tAugmentation recipe name ('r' to see available recipes):  "
+                    )
+                    if recipe == "r":
+                        print("\n\twordnet, embedding, charswap, eda\n")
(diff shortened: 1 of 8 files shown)


textattack/commands/augment.py
```diff
@@ -28,62 +28,101 @@ def run(self, args):
...
+                    break
+
+                if text == "c":
+                    print("recipe: ", end=" ")
+                    args.recipe = input()
+                    print("pct-words-to-swap: ", end=" ")
+                    args.pct_words_to_swap = float(input())
+                    print("transformations-per-example: ", end=" ")
+                    args.transformations_per_example = int(input())
+                    continue
```
jxmorris12, line 109: [1] this is nifty. Allowing users to change the arguments without restarting the program is a really good idea. Let's try to make it look a little more professional:
- at the beginning of this if statement, let's print an orienting message for the user. maybe: `print("Changing augmenter arguments...")`, or something like that
- then, can you print a tab before all of the words typed as arguments? like: `print("\trecipe: ")`

[2] rename the prompts to be a little more user-friendly: 
- `"recipe" -> "Augmentation recipe name"`
- `"pct-words-to-swap" -> "Percentage of words to swap"`
- `"transformations-per-example" -> "Transformations per input example"` 

[3] btw, do you need the `end=" "` argument provided to these print statements? I'd expect the thing to look good as-is (since you added a space to the end of the printed string already)

textattack/commands/augment.py
```diff
@@ -28,62 +28,101 @@ def run(self, args):
...
+                if not text:
+                    continue
+
+                print("Augmenting...")
+                print("----------------------------")
+
+                augmenter = eval(AUGMENTATION_RECIPE_NAMES[args.recipe])(
+                    pct_words_to_swap=args.pct_words_to_swap,
+                    transformations_per_example=args.transformations_per_example,
+                )
```
jxmorris12, line 120: I don't think this should happen on each iteration of the loop. Some augmenter `__init__` methods are expensive. For example, the `EmbeddingAugmenter` loads a full word embedding matrix upon initialization, which can be really slow, especially if you don't have a lot of RAM.

To avoid this, we can create the augmenter before the loop, and then re-initialize it only if the user types `"c"` (to change its arguments).

textattack/commands/augment.py
Hanyu-Liu-123: Thanks for the suggestions! I implemented these changes along with some others for a better visualization, like command spacings when they are printed out. 

When asking for `"Augmentation recipe name"`, I also included a condition in which if users type in "r", the list of available recipes will be printed. It could serve as a nice reminder (though there are only four of them). Do you think this could be helpful?

PR #227, opened 2020-07-26
Model agnosticism

jxmorris12:
- change ":" in command-line API to "|" (should hopefully make paths work on Windows)
- add back `distilbert-base-uncased-mr` model
- remove commented-out imports all over the place
- remove `try_model_predict` and other messy PyTorch model prediction guessing
- create `ModelWrapper` abstract class that models can implement to be used for attacks (this will allow us to add models from TensorFlow, sklearn, etc. with ease)

### DIFF
--- a/textattack/commands/attack/attack_args_helpers.py
+++ b/textattack/commands/attack/attack_args_helpers.py
@@ -303,9 +303,18 @@ def parse_model_from_args(args):
         model = textattack.shared.utils.load_textattack_model_from_path(
             args.model, model_path
         )
-        model = textattack.models.wrappers.PyTorchModelWrapper(
-            model, model.tokenizer, batch_size=args.model_batch_size
-        )
+        # Choose the approprate model wrapper (based on whether or not this is
+        # a HuggingFace model).
+        if isinstance(
+            model, textattack.models.helpers.BERTForClassification
+        ) or isinstance(model, textattack.models.helpers.T5ForTextToText):
+            model = textattack.models.wrappers.HuggingFaceModelWrapper(
+                model, model.tokenizer, batch_size=args.model_batch_size
+            )
+        else:
+            model = textattack.models.wrappers.PyTorchModelWrapper(
+                model, model.tokenizer, batch_size=args.model_batch_size
+            )
     elif args.model and os.path.exists(args.model):
         # Support loading TextAttack-trained models via just their folder path.
         # If `args.model` is a path/directory, let's assume it was a model
--- a/textattack/commands/attack/attack_args_helpers.py
+++ b/textattack/commands/attack/attack_args_helpers.py
@@ -118,15 +118,14 @@ def load_module_from_file(file_path):
     return module
 
 
-def parse_transformation_from_args(args, model):
-    # Transformations
+def parse_transformation_from_args(args, model_wrapper):
     transformation_name = args.transformation
     if ARGS_SPLIT_TOKEN in transformation_name:
         transformation_name, params = transformation_name.split(ARGS_SPLIT_TOKEN)
 
         if transformation_name in WHITE_BOX_TRANSFORMATION_CLASS_NAMES:
             transformation = eval(
-                f"{WHITE_BOX_TRANSFORMATION_CLASS_NAMES[transformation_name]}(model, {params})"
+                f"{WHITE_BOX_TRANSFORMATION_CLASS_NAMES[transformation_name]}(model_wrapper.model, {params})"
             )
         elif transformation_name in BLACK_BOX_TRANSFORMATION_CLASS_NAMES:
             transformation = eval(
@@ -137,7 +136,7 @@ def parse_transformation_from_args(args, model):
     else:
         if transformation_name in WHITE_BOX_TRANSFORMATION_CLASS_NAMES:
             transformation = eval(
-                f"{WHITE_BOX_TRANSFORMATION_CLASS_NAMES[transformation_name]}(model)"
+                f"{WHITE_BOX_TRANSFORMATION_CLASS_NAMES[transformation_name]}(model_wrapper.model)"
             )
         elif transformation_name in BLACK_BOX_TRANSFORMATION_CLASS_NAMES:
             transformation = eval(
@@ -149,7 +148,6 @@ def parse_transformation_from_args(args, model):
 
 
 def parse_goal_function_from_args(args, model):
-    # Goal Functions
     goal_function = args.goal_function
     if ARGS_SPLIT_TOKEN in goal_function:
         goal_function_name, params = goal_function.split(ARGS_SPLIT_TOKEN)
@@ -169,7 +167,6 @@ def parse_goal_function_from_args(args, model):
 
 
 def parse_constraints_from_args(args):
-    # Constraints
     if not args.constraints:
         return []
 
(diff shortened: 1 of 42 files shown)


textattack/models/wrappers/pytorch_model_wrapper.py
```diff
@@ -0,0 +1,45 @@
...
+
+import textattack
+
+from .model_wrapper import ModelWrapper
+
+
+class PyTorchModelWrapper(ModelWrapper):
+    """Loads a PyTorch model (`nn.Module`) and tokenizer."""
+
+    # TODO test training code before putting up PR
```
uvafan, line 11: Has this TODO been done?

textattack/models/wrappers/pytorch_model_wrapper.py
```diff
@@ -0,0 +1,45 @@
...
+                f"PyTorch model must be torch.nn.Module, got type {type(model)}"
+            )
+
+        self.model = model.to(textattack.shared.utils.device)
+        self.tokenizer = tokenizer
+        self.batch_size = batch_size
+
+    @property
+    def get_model_device(self):
+        if hasattr(self.model, "model"):
```
uvafan, line 24: can you add a comment explaining why you have to do this? Why do some models have model attributes?

textattack/shared/attack.py
```diff
@@ -166,6 +164,10 @@ def filter_transformations(
             current_text: The current ``AttackedText`` on which the transformation was applied.
             original_text: The original ``AttackedText`` from which the attack started.
         """
+        # Remove any occurences of current_text in transformed_texts
```
uvafan, line 167: In which cases do we need to do this? I feel like this should be up to the transformation not to do?

textattack/shared/attacked_text.py
```diff
@@ -346,7 +346,11 @@ def generate_new_attacked_text(self, new_words):
     @property
     def tokenizer_input(self):
         """The tuple of inputs to be passed to the tokenizer."""
-        return tuple(self._text_input.values())
+        input_tuple = tuple(self._text_input.values())
+        # Prefer to return a string instead of a tuple with a single value.
+        if len(input_tuple) == 1:
+            input_tuple = input_tuple[0]
```
uvafan, line 352: nit: I would change this to `return input_tuple[0]`.

textattack/shared/utils/misc.py
```diff
@@ -57,34 +56,41 @@ def html_table_from_rows(rows, title=None, header=None, style_dict=None):
...
+    model_cache_path = textattack.shared.utils.download_if_needed(model_path)
+    train_args_path = os.path.join(model_cache_path, "train_args.json")
+    if not os.path.exists(train_args_path):
+        textattack.shared.logger.warn(
+            f"train_args.json not found in model path {model_path}. Defaulting to 2 labels."
+        )
+        return 2
+    else:
+        args = json.loads(open(train_args_path).read())
+        return args["num_labels"]
```
uvafan, line 71: Should we include a default to 2 similar to above here if `num_labels` isn't a key?

textattack/shared/utils/tensor.py
```diff
@@ -1,119 +1,31 @@
+import numpy as np
 import torch
 
-import textattack
 
-# from textattack.shared import utils
+def batch_model_predict(model_predict, inputs, batch_size=32):
```
uvafan, line 5: Wow this seems much nicer now. I assume it still works same as before?

textattack/transformations/word_swap_gradient_based.py
```diff
@@ -18,8 +19,16 @@ class WordSwapGradientBased(Transformation):
...
     """
 
-    def __init__(self, model, top_n=1):
+    def __init__(self, model_wrapper, top_n=1):
+        # Unwrap model wrappers. Need raw model for gradient.
+        if not isinstance(model_wrapper, textattack.models.wrappers.ModelWrapper):
+            raise TypeError(f"Got invalid model wrapper type {type(model_wrapper)}")
+        model = model_wrapper.model
+        self.model_wrapper = model_wrapper
+        self.model = model
```
uvafan, line 28: nit: combine this and line 26 into `self.model = model_wrapper.model`

uvafan: Also maybe set a `self.tokenizer` so you only have to change it in one place if you change it again?

textattack/transformations/word_swap_inflections.py
```diff
@@ -35,7 +35,7 @@ def _get_replacement_words(self, word, word_part_of_speech):
 
     def _get_transformations(self, current_text, indices_to_modify):
         words = current_text.words
-        sentence = Sentence(current_text.text)
+        sentence = Sentence(" ".join(words))
```
uvafan, line 38: Why this change?

textattack/models/wrappers/pytorch_model_wrapper.py
jxmorris12: yes :)

textattack/shared/utils/tensor.py
jxmorris12: yep (all the tests pass anyway)

textattack/models/wrappers/pytorch_model_wrapper.py
jxmorris12: it's just a handy thing to have for pytorch models. I added it bc it was a helper that was already in the code. Since we don't use it, though, let's get rid of it.

textattack/shared/attack.py
jxmorris12: I think this is a good idea. It's a bug that's appeared maybe 4 times in 4 separate transformations (most recently, BERT masked-LM swap) – and identity transformations break the search methods

textattack/shared/utils/misc.py
jxmorris12: yes, good idea

textattack/transformations/word_swap_inflections.py
jxmorris12: FLAIR applies part-of-speech tags to punctuation and bits of punctuation. Without this, the part-of-speech tags wouldn't line up with `current_text.words` (and an error would be thrown in special cases)

textattack/transformations/word_swap_gradient_based.py
jxmorris12: good idea

textattack/commands/attack/attack_args_helpers.py
```diff
@@ -18,6 +18,10 @@
     WHITE_BOX_TRANSFORMATION_CLASS_NAMES,
 )
 
+# The split token allows users to optionally pass multiple arguments in a single
+# parameter by separating them with the split token.
+ARGS_SPLIT_TOKEN = "|"
```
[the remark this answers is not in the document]
jxmorris12: yeah, don't think we can do much more unfortunately

textattack/shared/attack.py
uvafan: Can we throw an error if there's a duplicate then? Rather than quietly filtering it out?

textattack/transformations/word_swap_inflections.py
uvafan: Hmm can it assign the right POS to words without punctuation though?

jxmorris12: yeah, this is a good concern, I shall put it on Asana

textattack/shared/attack.py
jxmorris12: done

PR #239, opened 2020-08-05
metrics 

jxmorris12:


textattack/metrics/metrics.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class ModelAccuracy(Metric):
+    key = "Model accuracy"
+
+    @staticmethod
+    def compute(attack_results):
+        total_attacks = len(attack_results)
+        skipped_attacks = TotalSkippedAttacks.compute(attack_results)
+        model_accuracy = (total_attacks - skipped_attacks) * 100.0 / (total_attacks)
```
uvafan, line 71: Not introduced by your changes but this actually seems like an unreliable way of computing accuracy, e.g. for targeted classification even if the model is wrong the goal might not be met yet if there are more than 2 classes, so the attack won't be skipped even though it's wrong.

uvafan: Actually this also applied for `AccuracyUnderAttack` to some extent, perhaps we should say that these only make sense for untargeted classification attacks.

textattack/metrics/metrics.py
```diff
@@ -0,0 +1,161 @@
...
+                    num_words_changed
+                    * 100.0
+                    / len(result.original_result.attacked_text.words)
+                )
+            else:
+                perturbed_word_percentage = 0
+            perturbed_word_percentages.append(perturbed_word_percentage)
+
+        perturbed_word_percentages = np.array(perturbed_word_percentages)
+        print("perturbed_word_percentages ->", perturbed_word_percentages)
```
uvafan, line 130: remove before merging

textattack/loggers/attack_log_manager.py
```diff
@@ -53,112 +85,42 @@ def flush(self):
...
+        #           and
+        #       `log_metrics`
+        # TODO: (maybe) in metrics.py:
+        #   develop a level of abstraction
+        #   that will automatically support percentages/
+        #   averages/etc of different precisions (# decimal
+        #   points) and stuff, and also hopefully reduce
+        #   repeat computations / reuse things better
+        #
+        # TODO: choose smarter default metrics for seq2seq models,
```
uvafan, line 109: How are we going to make the default vary based on model and attack type given this setup?

uvafan:
Another thought: it would be nice if metrics for each sample attacked were in the output csv (when `--enable-csv` toggled of course)

textattack/commands/attack/run_attack_single_threaded.py
```diff
@@ -135,23 +124,21 @@ def run(args, checkpoint=None):
...
-                    args, attack_log_manager, worklist, worklist_tail
+                    args, worklist, worklist_tail, attack_results=attack.attack_results,
                 )
                 new_checkpoint.save()
-                attack_log_manager.flush()
 
         pbar.close()
         print()
         # Enable summary stdout
         if args.disable_stdout:
```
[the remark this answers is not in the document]
jxmorris12: no

qiyanjun:
@ @tahmid-kazi Please read through this.

qiyanjun:
@tahmid-kazi can you help summarize?

jxmorris12:
@jinyongyoo Closing this since it's getting close to a year. Please reopen it when you're back to working on it! :)

PR #240, opened 2020-08-06
Attack recipe benchmarking

jxmorris12:
- change attack recipes to classes
   - besides being generally nicer, this paves the way for us to add auto benchmarking via `textattack attack benchmark`
- replace `|` with `^` as delimiter for command-line args, because `|` was a terrible choice
- add a test to make sure `textattack train` works with no errors (on a tiny dataset)
- add 'run this on colab' links to jupyter notebooks
- write some documentation


docs/index.rst
```diff
@@ -1,7 +1,31 @@
 TextAttack
-======================================
+==============
+
+Welcome to the documentation for TextAttack! You've come to the right place.
```
uvafan, line 4: This feels unnecessary

docs/index.rst
```diff
@@ -1,7 +1,31 @@
...
+
+What is TextAttack?
+----------------------
+`TextAttack <https://github.com/QData/TextAttack>`__ is a Python framework for adversarial attacks and data augmentation in NLP.
```
uvafan, line 8: and adversairal trianing?

textattack/attack_recipes/bae_garg_2019.py
```diff
@@ -28,85 +29,88 @@ def BAEGarg2019(model):
...
-    # to (Jin et al., 2019).
-    #
-    # • "If there are multiple tokens can cause C to misclassify S when they
-    # replace the mask, we choose the token which makes Sadv most similar to
-    # the original S based on the USE score."
-    # • "If no token causes misclassification, we choose the perturbation that
-    # decreases the prediction probability P(C(Sadv)=y) the most."
-    #
-    search_method = GreedyWordSwapWIR(wir_method="delete")
+        # "To ensure semantic similarity on introducing perturbations in the input
```
uvafan, line 60: Are these tabs that got added in by accident or smth?

textattack/attack_recipes/genetic_algorithm_alzantot_2018.py
```diff
@@ -13,62 +13,67 @@
...
-    goal_function = UntargetedClassification(model)
-    #
-    # Perform word substitution with a genetic algorithm.
-    #
-    search_method = GeneticAlgorithm(
-        pop_size=60, max_iters=20, post_crossover_check=False
-    )
 
-    return Attack(goal_function, constraints, transformation, search_method)
+        return Attack(goal_function, constraints, transformation, search_method)
```
uvafan, line 79: another tab?

uvafan:
Also one other thing to consider: we should probably add code at the top of notebooks for installing textattack at the top of the documentation notebooks since we're adding a "Run in Colab"

textattack/attack_recipes/bae_garg_2019.py
jxmorris12: they are nested within a class

PR #259, opened 2020-08-22
fix faulty cache misses

jinyongyoo:
If constraint cache size is too small, it's possible that some of the early hits would be evicted by new entries by `filter_transformation_uncached`.


jxmorris12:
thanks @jinyongyoo !!

PR #262, opened 2020-08-25
update 'textattack eval' and add tests

jxmorris12:


textattack/commands/eval_model/eval_model_command.py
```diff
@@ -96,8 +103,10 @@ def register_subcommand(main_parser: ArgumentParser):
         add_model_args(parser)
         add_dataset_args(parser)
 
+        parser.add_argument("--random-seed", default=1776, type=int)
```
jinyongyoo, line 106: Is the default same as the `random-seed` for attack commands? Its 1776 here but I think for attack it's `str_to_int("TEXTATTACK")` which is 765. Probably better to keep it the same unless there's a reason for changing it.

PR #267, opened 2020-09-01
update flair version & centralize POS tagging

jxmorris12:


textattack/shared/utils/strings.py
```diff
@@ -163,3 +163,34 @@ def color_text(text, color=None, method=None):
...
         return "[[" + text + "]]"
+
+
+_flair_pos_tagger = None
+
+
+def flair_tag(sentence):
+    """Tags a `Sentence` object using `flair` part-of-speech tagger."""
+    global _flair_pos_tagger
+    if not _flair_pos_tagger:
```
jinyongyoo, line 174: Good idea

PR #268, opened 2020-09-01
Add CheckList Recipe

Hanyu-Liu-123:
This recipe uses the "invariance testing" concept from the __CheckList__ paper, aiming to perturb the texts without changing the sentiment of the sentence

The following features are implemented from __CheckList__ to __TextAttack__:

- _Extension & Contraction_: "I'd like some tea" -> "I would like some tea"
- _Changing Numbers_: "I have five kids" -> " I have two kids"
- _Changing Location_: "New York is a good place" -> "London is a good place"
- _Changing Names_: "Justin Bieber is a star" -> "Ricky Bobby is a star"


textattack/transformations/word_swap_change_location.py
```diff
@@ -0,0 +1,100 @@
+import numpy as np
+
+from textattack.shared.data import NAMED_ENTITIES
+from textattack.transformations import Transformation
+
+
+def cluster_idx(idx_ls):
```
jxmorris12, line 7: Hi, ok, so this `cluster_idx` seems to work well, but luckily it's been solved for us already:

```
import more_itertools as mit
>>> data = [1, 2, 3, 5, 6, 7]
>>> [list(group) for group in mit.consecutive_groups(data)]
```

so I think you can replace calls to `cluster_idx` with that one-liner (`[list(group) for group in ...]`).

textattack/transformations/word_swap_change_number.py
```diff
@@ -0,0 +1,161 @@
...
+    output = []
+    for cluster in ls:
+        word = words[cluster[0]]
+        for idx in cluster[1:]:
+            word = " ".join([word, words[idx]])
+        output.append([cluster, word])
+    return output
+
+
+def cluster_idx(idx_ls):
```
jxmorris12, line 22: (see above comment about replacing `cluster_idx` with a call to `mit.consecutive_groups`)

textattack/transformations/word_swap_change_number.py
```diff
@@ -0,0 +1,161 @@
...
+            else:
+                output.append([idx])
+                list_pos += 1
+            prev = idx
+        return output
+
+
+class WordSwapChangeNumber(Transformation):
+    def __init__(self, max_change=1, n=3, **kwargs):
+        """A transformation that recognize numbers in sentence, and return
```
jxmorris12, line 45: typo: recognize -> recognizes, return -> returns

textattack/transformations/word_swap_change_number.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def _get_transformations(self, current_text, indices_to_modify):
+        words = current_text.words
+        num_idx = []
+        num_words = []
+
+        # find indexes of alphabetical words
+        for idx in indices_to_modify:
+            word = words[idx].lower()
+            for number in STR_NUM:
```
jxmorris12, line 63: This seems slow -- O(WN), where W is the length of `indices_to_modify`, and N is the length of `STR_NUM`. We should try and avoid double loops like this. Any ideas on how to speed things up?

textattack/transformations/word_swap_change_number.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # cluster adjacent indexes to get whole number
+        num_idx = cluster_idx(num_idx)
+        num_words += idx_to_words(num_idx, words)
+
+        # replace original numbers with new numbers
+        transformed_texts = []
+        for num_word in num_words:
+            idx = num_word[0]
+            word = num_word[1]
```
jxmorris12, line 83: In Python, you can make this more concise like so:

```
for (idx, word) in num_words:
```

So idx is the indices of `word` and word is potentially multiple words right?

textattack/transformations/word_swap_change_number.py
```diff
@@ -0,0 +1,161 @@
...
+            return self._alter_number(num)
+        else:
+            try:
+                num = w2n.word_to_num(word)
+                num_list = self._alter_number(num)
+                return [num2words(n) for n in num_list]
+            except ValueError:
+                return []
+
+    def _alter_number(self, num):
```
jxmorris12, line 113: So is the point here to replace a number with another random number? Please make that clear in the docstring :)

textattack/transformations/word_swap_contract.py
```diff
@@ -0,0 +1,94 @@
...
+        "you are": "you're",
+    }
+
+    def _get_transformations(self, current_text, indices_to_modify):
+        """Return all possible transformed sentences, each with one
+        contraction."""
+        transformed_texts = []
+
+        words = current_text.words
+        indices_to_modify = list(indices_to_modify)
```
jxmorris12, line 74: you should also sort indices_to_modify here:
```
indices_to_modify = sorted(indices_to_modify)
```

textattack/transformations/word_swap_contract.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # search for every 2-words combination in reverse_contraction_map
+        for idx in indices_to_modify[:-1]:
+            word = words[idx]
+
+            next_idx = indices_to_modify[indices_to_modify.index(idx) + 1]
+            next_word = words[next_idx]
```
jxmorris12, line 81: **Easy suggestion**

Instead of this `next_idx` thing, you can just do something like:

```
for i, word_idx in indices_to_modify[:-1]:
  word = words[word_idx]
  next_word_idx = indices_to_modify[i+1]
  next_word = words[next_word_idx]
```

And I think there's a weird case you're forgetting where words in `indices_to_modify` form a contraction, even though they're not actually consecutive words in a sentence. Like if the sentence was "You are all" and indices_to_modify was {0, 2} then you might try and replace "You all" with "you'll" even though the words are non-consecutive.

Anyway, I think you can get around this with a:

```
if next_word_idx != word_idx + 1:
  continue 
```

**Potentially better suggestion**

Another way to do it would be calling the cluster function (the thing you called `cluster_idx` in the other classes) on `indices_to_modify`. Then iterate through the clusters and, if they're length two, check to see if they are replaceable contractions. Just an idea, either route is good with me.

textattack/transformations/word_swap_contract.py
```diff
@@ -0,0 +1,94 @@
...
+        "who will": "who'll",
+        "who is": "who's",
+        "who have": "who've",
+        "why is": "why's",
+        "will not": "won't",
+        "would not": "wouldn't",
+        "you would": "you'd",
+        "you will": "you'll",
+        "you are": "you're",
+    }
```
jxmorris12, line 66: since this is an exact reversal of contraction map, I think you should move contraction map to the `data.py` file and then define an __init__ function here that does something like:
```
self.reverse_contraction_map = {v: k for k, v in CONTRACTION_MAP.items()}
```
(where CONTRACTION_MAP is imported from the data file)

That should automatically reverse the contraction map from data.py. Then if we add/remove a contraction it'll automatically update this reverse contraction map too!

textattack/transformations/word_swap_extend.py
```diff
@@ -0,0 +1,107 @@
...
+        """Return all possible transformed sentences, each with one
+        extension."""
+        transformed_texts = []
+        words = current_text.words
+        for idx in indices_to_modify:
+            word = words[idx]
+            # expend when word in map
+            if word in self.contraction_map:
+                expanded = self.contraction_map[word].split()
+                print(word, expanded)
```
jxmorris12, line 98: pls remove this print statement

textattack/transformations/word_swap_extend.py
```diff
@@ -0,0 +1,107 @@
...
+                transformed_text = current_text.replace_word_at_index(idx, expanded[0])
```
jxmorris12, line 99: I think you should just be able to do:
```
transformed_text = current_text.replace_word_at_index(idx, expanded)
```

and remove the calls to `insert_text_after_word_index`. `AttackedText` should be able to handle multi-word replacements like this; let me know if I'm wrong.

textattack/transformations/word_swap_change_number.py
```diff
@@ -0,0 +1,161 @@
...
+            word = num_word[1]
+            replacement_words = self._get_new_number(word)
+            for r in replacement_words:
+                if r == word:
+                    continue
+                text = current_text.replace_word_at_index(idx[0], str(r))
+                if len(idx) > 1:
+                    index = idx[1]
+                    for i in idx[1:]:
+                        text = text.delete_word_at_index(index)
```
jxmorris12, line 92: I've seen this pattern in a couple places: to replace multiple words with a single word, you replace the first word, then delete the rest. That's a good solution. However, I think you can save some code by using the `replace_words_at_indices` method from AttackedText (in all places where you do this, including here!).

In this case, I think it would look like:
```text = current_text.replace_words_at_indices(idx, str(r))```

File: textattack/shared/attacked_text.py [hunk 13]
```diff
@@ -146,6 +147,31 @@ def pos_of_word_index(self, desired_word_idx):
...
+            if word_idx == desired_word_idx:
+                return flair_ner_list[word_idx_in_flair_tags]
+            else:
+                flair_word_list = flair_word_list[word_idx_in_flair_tags + 1 :]
+                flair_ner_list = flair_ner_list[word_idx_in_flair_tags + 1 :]
+
+        raise ValueError(
+            f"Did not find word from index {desired_word_idx} in flair POS tag"
+        )
+
```
jxmorris12, line 174: Awesome work! This was the hardest part of everything, I think, and looks like you nailed it!

jxmorris12 (changes requested):
This is awesome. The results that I've found are really funny/interesting.

I have one major overarching question: how do these transformations (extend, contract, changename, changenumber, changelocation) handle uppercased words? For example, if a location is lowercased (like 'new york') what happens then? What about if a contraction is uppercased ('You all')?

Overall great work! Can't wait to get this merged.

textattack/transformations/word_swap_change_number.py
Hanyu-Liu-123: The problem with using `replace_words_at_indices` is that we can only replace n idx with exactly n words, or it's going to give this error: ` f"Cannot replace {len(new_words)} words at {len(indices)} indices."`

Should we change `replace_words_at_indices ` to accompany that?

textattack/transformations/word_swap_change_number.py
Hanyu-Liu-123: I thought about doing something like:
```
for idx in indices_to_modify:

      word = words[idx].lower()

      if word in STR_NUM:

            ............................
```

But this will miss some of the number combinations like "thirty-five" since we don't have them in `STR_NUM`

Do you think this would be a worthy trade? Maybe there are better options we can think of.

textattack/transformations/word_swap_change_number.py
jxmorris12: go for it 👍

textattack/transformations/word_swap_change_number.py
jxmorris12: ok, understood – I think it's fine to leave as is for now 👍

jxmorris12 (changes requested):
just change the replace_words_at_idxs thing -- (1) remove the error and then (2) use this function instead of a replacement and multiple insertions!

gaguilar:
Very excited to see this PR coming to the library! 🙃

jxmorris12 (approved):

PR #279, opened 2020-09-17
resolve the issue that passing an empty list to dataset would attack the entire dataset

MintForever:


jxmorris12 (approved):
awesome work!

jxmorris12 (changes requested):
looks like we need to fix one more thing

PR #281, opened 2020-09-25
Update readme with CheckList recipes

Hanyu-Liu-123:


jinyongyoo (approved):

jinyongyoo (changes requested):
Misspelling "substitues". Should it say "substitutes"?

Also, I think it should be `automodule:: textattack.attack_recipes.checklist_ribeiro_2020`

jinyongyoo (approved):

PR #284, opened 2020-09-29
CheckList-Inv augmentation

Hanyu-Liu-123:
This augmentation recipe uses INV testing transformations implemented in the CheckList attack recipe

### DIFF
--- a/textattack/augmentation/recipes.py
+++ b/textattack/augmentation/recipes.py
@@ -145,6 +145,18 @@ class CharSwapAugmenter(Augmenter):
 
 
 class CheckListAugmenter(Augmenter):
+    """Augments words by using the transformation methods provided by CheckList
+    INV testing, which combines:
+
+    - Name Replacement
+    - Location Replacement
+    - Number Alteration
+    - Contraction/Extension
+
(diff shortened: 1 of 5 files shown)


textattack/augmentation/recipes.py
```diff
@@ -142,3 +142,41 @@ def __init__(self, **kwargs):
...
+            [
+                WordSwapChangeNumber(),
+                WordSwapChangeLocation(),
+                WordSwapChangeName(),
+                WordSwapExtend(),
+                WordSwapContract(),
+            ]
+        )
+
+        constraints = [DEFAULT_CONSTRAINTS[0]]
```
jxmorris12, line 180: Do you think it makes sense to include both pretransformation constraints in this case?

jxmorris12 (approved):

textattack/augmentation/recipes.py
Hanyu-Liu-123: I think if we include `StopwordModification` as constraints, `WordSwapContract` will be handicapped because words like "I" or "you" are stop words. So when we try to contact "I will" to "I'll", `WordSwapContract` will be stopped by `StopwordModification`

jxmorris12: Ah good call!

jxmorris12:
Great work Hanyu!!

PR #287, opened 2020-10-01
Update README.md

Hanyu-Liu-123:
There is a small mistake in the `Augmentation` session of README. The example code gives `--pct-words-to-swap 4 `, which will means altering 400% of the original example and will cause an error. I changed it to 0.1 to match the description of that example.


jxmorris12 (approved):

jxmorris12:
thank u

PR #291, opened 2020-10-02
Add stanza support for part-of-speech constraint

k-ivey:
* Added Stanza support for the `PartOfSpeech` constraint
* Added test for using Stanza as the part of speech tagger


jxmorris12 (approved):
nice work!!!

PR #292, opened 2020-10-05
Add model-agnostic gradient retreival + gradient-based GreedyWordWIR search

jinyongyoo:
Changes:

- Add `get_grad` method for model wrappers. This method takes in input string and returns gradient w.r.t. each input token.
- Update `WordSwapGradientBased` to work with the new `get_grad` method
- Add property to search methods indicating whether they are white-box or black-box search method.
- Make the model wrapper available to white-box search methods in `attack.py`
- Remove `BERTForClassification` class. Since all of our models have moved to HuggingFace, we never instantiate this class. 
- Raise TF logging level to make it even less verbose.
- Add new tests


textattack/models/wrappers/huggingface_model_wrapper.py
```diff
@@ -24,32 +47,71 @@ def __call__(self, text_input_list):
...
+            raise TypeError(
+                f"{type(self.model)} class does not take in `labels` to calculate loss. "
+                "One cause for this might be if you instatiated your model using `transformer.AutoModel` "
+                "(instead of `transformers.AutoModelForSequenceClassification`)."
+            )
+
+        loss.backward()
+
+        # grad w.r.t to word embeddings
+        grad = emb_grads[0][0].cpu().numpy()
```
jxmorris12, line 109: sorry, why is this `emb_grads[0][0]` instead of `emb_grads[0]`? I haven't used Hooks in awhile.

textattack/models/wrappers/pytorch_model_wrapper.py
```diff
@@ -8,14 +9,15 @@
 class PyTorchModelWrapper(ModelWrapper):
     """Loads a PyTorch model (`nn.Module`) and tokenizer."""
 
-    def __init__(self, model, tokenizer, batch_size=32):
+    def __init__(self, model, tokenizer, loss=CrossEntropyLoss(), batch_size=32):
```
jxmorris12, line 12: Can you add `loss` to the docstring? (This docstring was already kind of incomplete, sorry)

textattack/search_methods/greedy_word_swap_wir.py
```diff
@@ -69,12 +71,47 @@ def _get_index_order(self, initial_text):
...
+                    token = grad_output["tokens"][j].lower()
+                    # remove "##" if it's a subword
+                    token = token.replace("##", "")
+                    idx = word.find(token)
+                    if idx == 0:
+                        word = word[idx + len(token) :]
+                        matched_tokens.append(j)
+                        a.append(token)
+                        last_matched = j
+                    j += 1
```
jxmorris12, line 103: This is my biggest comment: I think that this model-specific string processing should happen in the model wrappers. 

Otherwise, we'll find ourselves updating this bit of code every time we add a model with different tokenization specifics (for example, a char-level model, or a model that marked subwords with ** instead of ##). What do you think?

textattack/search_methods/greedy_word_swap_wir.py
```diff
@@ -69,12 +71,47 @@ def _get_index_order(self, initial_text):
...
+
+                if not matched_tokens:
+                    # Reset j to most recent match
+                    j = last_matched
+                    index_scores[i] = 0.0
+                else:
+                    agg_grad = np.mean(gradient[matched_tokens], axis=0)
+                    index_scores[i] = np.linalg.norm(agg_grad, ord=1)
```
jxmorris12, line 111: So this is the sum of the absolute value of the gradients at each position? Is this how they do it in a paper? Just wondering. If so we might want to cite.

textattack/search_methods/particle_swarm_optimization.py
```diff
@@ -40,6 +40,7 @@ def __init__(
         self.pop_size = pop_size
         self.post_turn_check = post_turn_check
         self.max_turn_retries = 20
+        self.is_black_box = True
```
jxmorris12, line 43: Do you need this and the @property below? I think you can leave this and remove the @property methods, if I recall correctly.

textattack/shared/attack.py
```diff
@@ -55,7 +55,9 @@ def __init__(
         self.transformation = transformation
         if not self.transformation:
             raise NameError("Cannot instantiate attack without transformation")
-        self.is_black_box = getattr(transformation, "is_black_box", True)
+        self.is_black_box = (
+            getattr(transformation, "is_black_box", True) or search_method.is_black_box
```
jxmorris12, line 59: I think this should be an and, not an or. Because an attack with a white-box search method and black-box transformation is still a "white-box attack", right?

jxmorris12 (changes requested):
Looks great! This is a big improvement!

textattack/shared/attack.py
jinyongyoo: Oh yeah you're right. Good catch.

textattack/search_methods/particle_swarm_optimization.py
jinyongyoo: Yeah I was debating whether to set it as regular attribute or @property, and decided to go with @property and forgot to delete it. I think we should just keep it as @property since that's how I do it in the parent `SearchMethod` class.

textattack/search_methods/greedy_word_swap_wir.py
jinyongyoo: I'm assuming you're talking about the HotFlip paper. But I thought HotFlip does not actually perform this type of greedy-word-wir type search? They do beam search. I don't recall any paper actually doing this. 

As for the L1 norm instead of L2 norm, I wasn't sure actually which one to pick and it wasn't clear from reading the papers. But I do know that [AllenInterpret](https://github.com/allenai/allennlp/blob/7c10fd26d0833e6f97f8559576ad759f4366a234/allennlp/interpret/saliency_interpreters/simple_gradient.py#L44) does L1 norm, so that's why I also chose it.

textattack/search_methods/greedy_word_swap_wir.py
jinyongyoo: I agree.

tests/sample_outputs/run_attack_hotflip_lstm_mr_4.txt
```diff
@@ -26,7 +26,7 @@
       )
     (3): RepeatModification
     (4): StopwordModification
-  (is_black_box):  False
```
jinyongyoo, line 29: Change this back to `False`

textattack/models/wrappers/huggingface_model_wrapper.py
jinyongyoo: I believe it's because of input technically a batch of size 1. But I'll double check.

jinyongyoo: Confirmed that it's because of batch size.

textattack/search_methods/greedy_word_swap_wir.py
jinyongyoo: I thought about this more. I think this also works with char-level tokenization since we can treat characters as subwords too. Now regarding different subword prefixes, I'm not sure there's a good way to resolve this since it's something that depends on the model, and doing it in the model wrapper is not gonna be resolve this since model wrapper doesn't know what prefix model uses. 

To address this however, I moved the alignment step to `AttackedText` and moved tokenization + stripping of prefixes to model wrapper.

jxmorris12: Looks great!

tests/test_command_line/test_attack.py
```diff
@@ -173,6 +183,7 @@ def test_command_line_attack(name, command, sample_output_file):
     # / is escaped in python 3.6, but not 3.7+, so we support both
     desired_re = (
         re.escape(desired_output)
+        .replace("/\\.\\/", ".")
```
jxmorris12, line 186: Nice lol

jxmorris12 (approved):
This is huge!

PR #293, opened 2020-10-06
have flair use textattack tokenization method

jinyongyoo:
Flair's tokenizer sometimes causes issues with POS tagging since it doesn't tokenize a sentence into words like TextAttack does. For example, if we have the word "KO'ed", Flair will tokenize it as "KO'", "ed", but TextAttack treats "KO'ed" as one word. So when we perform POS tagging, we won't find "KO'ed" in the word list returned by Flair.

Flair allows users to specify their own tokenizer as a function that takes in a string and returns a list of tokens. `textattack.shared.utils.words_from_text` does this exactly.


Hanyu-Liu-123 (approved):
Looks good to me :-)

jxmorris12 (approved):

PR #297, opened 2020-10-09
Add a CoLA based constraint

k-ivey:
- Add a constraint to check that the attacked text is deemed grammatically acceptable by a model pre-trained on the CoLA dataset
- I'm not convinced that `COLA` is the best name for this constraint, but I couldn't come up with anything else. Suggestions are welcome. 
- The `run_attack_targeted_mnli_misc` test is failing on my local machine with  "Avg num queries: 182.0" rather than the new 181.0 value. I was unable to determine the issue since none of my changes should have affected this test.

### DIFF
--- a/tests/sample_outputs/run_attack_cnn_cola.txt
+++ b/tests/sample_outputs/run_attack_cnn_cola.txt
@@ -1,0 +1,59 @@
+/.*/Attack(
+  (search_method): GreedyWordSwapWIR(
+    (wir_method):  unk
+  )
+  (goal_function):  UntargetedClassification
+  (transformation):  WordSwapWordNet
+  (constraints): 
+    (0): COLA(
+        (max_diff):  0.1
+        (model_name):  textattack/bert-base-uncased-CoLA
+        (compare_against_original):  True
+      )
+    (1): BERTScore(
+        (min_bert_score):  0.7
+        (model):  bert-base-uncased
+        (score_type):  f1
+        (compare_against_original):  True
+      )
+  (is_black_box):  True
+) 
+/.*/
+--------------------------------------------- Result 1 ---------------------------------------------
+[92m2 (72%)[0m --> [91m1 (53%)[0m
+
+Contrary to other reviews, I have zero complaints about the service or the prices. I have been getting tire service here for the past 5 years now, and compared to my experience with places like Pep Boys, these [92mguys[0m are experienced and know what they're doing. \nAlso, this is one place that I do not feel like I am being taken advantage of, just because of my gender. Other auto mechanics have been notorious for capitalizing on my ignorance of cars, and have sucked my bank account dry. But here, my service and road coverage has all been well explained - and let up to me to decide. \nAnd they just renovated the waiting room. It looks a lot better than it did in previous years.
+
+Contrary to other reviews, I have zero complaints about the service or the prices. I have been getting tire service here for the past 5 years now, and compared to my experience with places like Pep Boys, these [91mblackguard[0m are experienced and know what they're doing. \nAlso, this is one place that I do not feel like I am being taken advantage of, just because of my gender. Other auto mechanics have been notorious for capitalizing on my ignorance of cars, and have sucked my bank account dry. But here, my service and road coverage has all been well explained - and let up to me to decide. \nAnd they just renovated the waiting room. It looks a lot better than it did in previous years.
+
+
+--------------------------------------------- Result 2 ---------------------------------------------
+[91m1 (61%)[0m --> [92m2 (51%)[0m
+
+Last summer I had an appointment to get new tires and had to wait a super long time. I also went in this week for them to fix a minor problem with a tire they put on. They \""fixed\"" it for free, and the very next morning I had the same issue. I called to complain, and the \""manager\"" didn't even apologize!!! So frustrated. Never going back.  They [91mseem[0m overpriced, too.
+
+Last summer I had an appointment to get new tires and had to wait a super long time. I also went in this week for them to fix a minor problem with a tire they put on. They \""fixed\"" it for free, and the very next morning I had the same issue. I called to complain, and the \""manager\"" didn't even apologize!!! So frustrated. Never going back.  They [92mlook[0m overpriced, too.
+
+
+--------------------------------------------- Result 3 ---------------------------------------------
+[92m2 (76%)[0m --> [91m1 (63%)[0m
+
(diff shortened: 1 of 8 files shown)


tests/sample_outputs/run_attack_cnn_cola.txt
```diff
@@ -0,0 +1,59 @@
+Attack(
```
jinyongyoo, line 1: Really minor thing, but when creating new tests, it's a good idea to add `/.*/` before "Attack" in line 1 (so like `/.*/Attack(` and another one between the printing of attack and `---Result 1---` (e.g. line 21). This is because sometimes, extra log output might be captured in these spots and cause test failures even though the main content are correct.

textattack/constraints/grammaticality/cola.py
```diff
@@ -0,0 +1,79 @@
...
+        compare_against_original=True,
+    ):
+        super().__init__(compare_against_original)
+        if not isinstance(max_diff, float):
+            raise TypeError("max_diff must be a float")
+        if max_diff < 0.0 or max_diff > 1.0:
+            raise ValueError("max_diff must be a value between 0.0 and 1.0")
+        if model_name not in HUGGINGFACE_DATASET_BY_MODEL:
+            raise ValueError(
+                "model_name must be the name of a pre-trained model on the CoLA dataset"
```
jinyongyoo, line 42: One small suggestion for error message is that we might want to tell the user what they supplied for `model_name`. So like `f"model_name {model_name} must be ....`

textattack/constraints/grammaticality/cola.py
```diff
@@ -0,0 +1,79 @@
...
+            )
+
+        self.max_diff = max_diff
+        self.model_name = model_name
+        self._reference_score_cache = lru.LRU(2 ** 10)
```
jinyongyoo, line 47: Good use of cache!

jinyongyoo (commented):
Good work! I like the use of cache, detailed comments, and overall completeness. I do want to raise two concerns. First, is it possible for user to use their own model? We don't have to cover every possibility, but limiting it to ones we have defined in `HUGGINGFACE_DATASET_BY_MODEL` seems too restrictive. We can simply change `model_name` to actually be the value you put into `AutoModelForSequenceClassification.from_pretrained(...)` and change the default `model_name` to be `textattack/bert-base-uncased-cola`.

Secondly, it seems `max_diff` will be heavily dependent on the number of sentences we have. For instance, suppose our text samples mostly have one sentence (like SST-2). Then, we will frequently have cases where reference score is 1 and score for transformed text is 0 since there's only one sentence. Therefore, `max_diff` will usually have to be 1. But for longer text like Yelp or IMDB, `max_diff` might actually have to be some small value. 

This will become an issue when the dataset contains samples with varying length. Can we perhaps scale `max_diff` by number of sentences?

textattack/constraints/grammaticality/cola.py
k-ivey: Good catch, but by allowing users to supply their own model, I don't think we'll be able to have a check like this anymore.

k-ivey:
I'll add support for users to use their own models from the huggingface model hub - this will be a simple change like you said. 

The issue with `max_diff` is a bit more complicated. If the text only has one sentence, then a `max_diff` of 1 would cause any piece of text to automatically pass the check, making the constraint useless. Similarly if `max_diff` is 0.2, then only text with 5+ sentences could actually have a change in the number of valid sentences to pass the check. So for datasets with differing length texts, we would need some sort of inverse scale for smaller pieces of text to have sentences become invalid when attacked and still pass the check. Another option is to change the `max_diff` parameter to be the maximum difference in the number of valid sentences between the reference and transformed text. However, this has it's own issues since it doesn't scale at all with the number of sentences. 

I think the first/ current way may be better. For text with less sentences, fewer sentences could be made invalid when attacked; while for text with more sentences, more sentences could be made invalid. This matches my intuition of how the constraint would work. Like you said, an issue would arise where the number of sentences and the `max_diff` parameter are small and the user wants small pieces of text to have sentences become invalid when attacked yet still pass the check, but I think this somewhat goes against the purpose of this constraint. Combining the two methods and having an absolute difference and a relative difference (e.g. the transformed text and reference text can a relative difference of 0.2 or an absolute difference of 1 sentence) would make the constraint more complicated, but would allow users to have this functionality. 

Do you have any thoughts on the two methods? Or an idea of a scale we could use to allow smaller pieces of text to have more invalid sentences without breaking the constraint for larger text?

textattack/constraints/grammaticality/cola.py
jinyongyoo: Yes.

jinyongyoo:
Seems like I made a mistake. What I actually meant was that `max_diff` would actually have to 0 for single-sentence cases.
You’re right that `max_diff=1` would make the constraint useless. Also, I can see that setting `max_diff` to be like 0.2 would force all text with less than 5 sentences to have all valid sentences (since differing by 1 would result in percent difference being greater than 0.2). 

I think, however, the fact that I got confused up with this all indicates that other users will also have trouble understanding how to set `max_diff`. It's very natural to think of the difference in just the absolute number of sentences instead of relative difference. My guess (might be wrong) is that most users would actually just use absolute difference instead of trying to deal with the relative difference since the former is clearer. 

What I would suggest is that we provide the option for both absolute and relative difference. One way to do this would be so that if `max_diff` is an integer, it would be absolute difference, and if it's a float, it would relative difference.

@jxmorris12 Do you have other suggestions for this?

textattack/constraints/grammaticality/cola.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+    def __init__(
+        self,
+        max_diff,
+        model_name="textattack/bert-base-uncased-CoLA",
+        compare_against_original=True,
+    ):
+        super().__init__(compare_against_original)
+        if not isinstance(max_diff, float):
```
jinyongyoo, line 34: Values like 2 is an int and not a float in Python, so we would fail this check. I think should check if `max_diff` is float or int, and base on that, determine how to calculate the threshold value for `num_valid` in `_check_constraint`. So like
``` 
if isinstance(self.max_diff, int):
      # set absolute threshold value
else:
      # set relative threshold value

# check if num_valid < threshold
```

jinyongyoo (changes requested):

jinyongyoo (approved):
LGTM!

jxmorris12:
This is great! Nice work @k-ivey

PR #298, opened 2020-10-09
New augmenter

Hanyu-Liu-123:
The augmenter we have right now only supports single word swaps, making many transformations inapplicable for augmentation purposes. 

The complicity here is due to the parameter `pct_words_to_swap `, which allows users to set the maximum number of words to swap during augmentation. We have to call transformation as many times as we can, but the max number of words swapped can not exceed the allowance set by user. What I changed here is to use the length of `AttackedText.attack_attrs["modified_indices"]` to count the number of swapped words, instead of the previous method of using a list of random indices to perform single word swap. In this way, we could perform any transformation, see if it fits the constraints/allowance, then decide to whether return the augmentation or continue swapping.

An option to display current Augmenter parameter is added to interactively running the augmenter.

I also noticed that we have only augmentation test! I'll add more if it won't slow down the current test too much.

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -16,3 +16,29 @@ def test_embedding_augmenter():
         "There is nothing either good or unfavourable, but thinking makes it so."
     )
     assert augmented_s in augmented_text_list
+
+
+def test_checklist_augmenter():
+    from textattack.augmentation import CheckListAugmenter
+
+    augmenter = CheckListAugmenter(pct_words_to_swap=0.01, transformations_per_example=64)
+    s = "I'll be happy to assist you."
+    augmented_text_list = augmenter.augment(s)
+    augmented_s = "I will be happy to assist you."
+    assert augmented_s in augmented_text_list
+
+    s = "I will be happy to assist you."
+    augmented_text_list = augmenter.augment(s)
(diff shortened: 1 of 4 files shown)


jxmorris12 (approved):
LGTM! I am pleasantly surprised that this didn't break our other augmentation test. (Also, a little confused!)

Do you mind adding a test that programmatically calls the augmenter and verifies it returns some correct outputs? It shouldn't make the tests much slower, since the augmenter runs very quickly on the CPU. You can make sure to pick a transformation that will run quickly.

Hanyu-Liu-123:
> LGTM! I am pleasantly surprised that this didn't break our other augmentation test. (Also, a little confused!)
> 
> Do you mind adding a test that programmatically calls the augmenter and verifies it returns some correct outputs? It shouldn't make the tests much slower, since the augmenter runs very quickly on the CPU. You can make sure to pick a transformation that will run quickly.

I'll definitely add more tests for augmentation. I think the reason why it didn't break any tests is because we only have 1 augmentation test that actually checks for the outputs. But fortunately it passed that one test we have right now.

jxmorris12:
Amazing! Great work again @Hanyu-Liu-123

PR #299, opened 2020-10-11
Added a new Constrain MultilingualUniversalSentenceEncoder

FlorianO:
This is just a tiny Pull Request to add Multilingual Universal Sentence Encoder (an extension of Universal Sentence Encoder).

See https://tfhub.dev/google/universal-sentence-encoder-multilingual/3 for details.


requirements.txt
```diff
@@ -17,6 +17,7 @@ torch
 transformers==3.3.0
 tensorflow>=2
 tensorflow_hub
+tensorflow_text>=2
```
jinyongyoo, line 20: Since `tensorflow_text` is not needed, could you remove this from the requirements.txt?

textattack/constraints/semantics/sentence_encoders/multilingual_universal_sentence_encoder/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .multilingual_universal_sentence_encoder import (
```
jinyongyoo, line 1: I'm not sure if it's necessary to create new a directory for MUSE. Since MUSE is an extension of USE and since we already have a directory for USE, it seems to make more sense to just have both MUSE and USE under `textattack/constraints/semantics/sentence_encoders/universal_sentence_encoder`.

jinyongyoo (changes requested):
First off, thank you for adding this new constraint! Happy to see contributions from the community! Overall, the PR looks good. I have left two comments that I think you should look at.

requirements.txt
FlorianO: If you remove it there will be problems with Tensorflow:
`Op type not registered 'SentencepieceOp'`
See https://github.com/tensorflow/tensorflow/issues/38597 for an example

textattack/constraints/semantics/sentence_encoders/multilingual_universal_sentence_encoder/__init__.py
FlorianO: I'm fine with this as well, how would you distinguish between them when calling on the command line?
As another argument?

FlorianO: Ah, got your point. Sure i can move it to the other directory. :)

requirements.txt
jinyongyoo: Yes I just also realized this while trying to run an attack with MUSE. Please ignore the previous comment.

jinyongyoo (commented):
@FlorianO, it seems like `import tensorflow_text` is needed in `multilingual_universal_sentence_encoder.py` to not get the `Op type not registered 'SentencepieceOp'` error (https://github.com/tensorflow/hub/issues/463). 

You can have `flake8` ignore the line and not raise "imported but no used" error by adding `# noqa: F401` next to the import statement (e.g. `import tensorflow_text  # noqa: F401`)

textattack/constraints/semantics/sentence_encoders/multilingual_universal_sentence_encoder/__init__.py
FlorianO: Fixed it.

FlorianO:
Everything should be fixed now, thank you for reviewing! :)

jinyongyoo (approved):

PR #301, opened 2020-10-13
Fix typo in README.md

madaan:
The sample augmentation command currently uses ``--pct-words-to-swap 4``, which will trigger ``AssertionError: pct_words_to_swap must be in [0., 1.]''.
This commit fixes it to .1 as intended.


jxmorris12 (approved):

jxmorris12:
thank you!!

PR #303, opened 2020-10-15
Fix for tut0

sanchit97:
Minimal changes for files blocking successful run of tut0


textattack/commands/eval_model/eval_model_command.py
```diff
@@ -32,7 +32,7 @@ class EvalModelCommand(TextAttackCommand):
 
     def get_preds(self, model, inputs):
         with torch.no_grad():
-            preds = textattack.shared.utils.batch_model_predict(model, inputs)
+            preds = model(inputs)
```
jxmorris12, line 35: Is there any reason why you changed this line? 

You're correct that `inputs` here is already batched according to `args.batch_size`, but I think we want to keep some of the [logic from batch_model_predict](https://github.com/QData/TextAttack/blob/master/textattack/shared/utils/tensor.py#L15-L27). 

I think the best way to do it would be to remove the loop in `eval` and just get all the predictions at once using:
```python
with torch.no_grad():
  preds = textattack.shared.utils.batch_model_predict(model, inputs, batch_size=args.batch_size)
```

but there's no reason you should have to change that in your PR.

textattack/shared/utils/tensor.py
```diff
@@ -11,7 +11,7 @@ def batch_model_predict(model_predict, inputs, batch_size=32):
     i = 0
     while i < len(inputs):
         batch = inputs[i : i + batch_size]
-        batch_preds = model_predict(batch)
+        batch_preds = model_predict(batch)[0]
```
jxmorris12, line 14: Please revert this as well. This tuple unwrapping happens in `HuggingFaceModelWrapper`: https://github.com/QData/TextAttack/blob/master/textattack/models/wrappers/huggingface_model_wrapper.py#L48

textattack/models/wrappers/pytorch_model_wrapper.py
```diff
@@ -21,6 +21,7 @@ def __init__(self, model, tokenizer, batch_size=32):
     def __call__(self, text_input_list):
         model_device = next(self.model.parameters()).device
         ids = self.tokenize(text_input_list)
+        ids = [i["input_ids"] for i in ids]
```
jxmorris12, line 24: Please revert this line. `"input_ids"` is a key that belongs to the inputs of models that use Hugging Face tokenizers.  So I'm guessing that the model is being imported from the file using the PytorchModelWrapper when it's supposed to be a HuggingFaceModelWrapper. I think if you fixed that then this 'issue' as well as the one in `tensor.py` should fix itself.

jxmorris12 (changes requested):
Hi, nice work fixing the bug! Unfortunately it's also caused some problems with other parts of the framework, which is why the tests fail.

This specific issue regards how models saved via `textattack train` are 'automatically' loaded and used for `textattack eval` or `textattack attack`. So I think the only changes you should have to make will be to our argparsing logic-- that is, files within `textattack/commands`. If the model is correctly loaded, then the other parts of TextAttack should run error-free.

In this case, you're totally right that we needed to pass `model.model` to the model wrapper constructor; however, I think that it should be a `HuggingFaceModelWrapper`, not a `PytorchModelWrapper`. The rest of this stuff (like passing tokens in the right format) is handled in `HuggingFaceModelWrapper`. So if you can get the model read in as the correct class, the `textattack eval` line should work properly!

textattack/models/wrappers/pytorch_model_wrapper.py
sanchit97: Gotcha, Reverted

textattack/shared/utils/tensor.py
sanchit97: Reverted

textattack/commands/eval_model/eval_model_command.py
sanchit97: Seems like a totally separate discussion. Reverted.

qiyanjun:
When @sanchit97 used HuggingFaceModelWrapper, everything works fine. So pull request closed for this.

jxmorris12:
@qiyanjun isn't the command `textattack attack --model /path/to/model` still broken?

sanchit97:
Yes, I have made only the one changes we discussed before and reverted all. Do I create a new PR for that or we re-open this?

jxmorris12:
Either approach works @sanchit97! The easiest is probably to push to this branch and re-open the pull request.

qiyanjun:
I see. Please reopen. I misunderstood your message.

qiyanjun:
@sanchit97 have you added the change? 

btw: please also pay attention to https://github.com/QData/TextAttack/pull/322

sanchit97:
Yes, I have.

jxmorris12:
@jinyongyoo  do you know why the tests are failing?

jinyongyoo:
I'm somewhat confused by this PR. Why are we unwrapping model wrapper returned from `model_from_args` and then wrapping it again as `HuggingFaceModelWrapper`? [`model_from_args`](https://github.com/QData/TextAttack/blob/41e747215bb0f01c511af95b16b94704c780cd5a/textattack/commands/train_model/train_args_helpers.py#L90) already wraps the model and tokenizer in appropriate model wrapper.

jinyongyoo:
Also, I reran the tests and it seems like the failure with `bert-base-uncased-MNLI` is a temporary one due to network connection. But it is failing CheckList augmenter test due to reason that we are not responsible for. Hanyu actually fixed last week by fixing torch version to `1.6.0`. Please rebase from master to resolve this.

qiyanjun:
@sanchit97 can you rebase?

qiyanjun:
@sanchit97 please rebase from master!  https://textattack.readthedocs.io/en/latest/1start/support.html#start-contributing-pull-requests

qiyanjun:
@sanchit97 please run "black textattack"

jinyongyoo (approved):

PR #304, opened 2020-10-16
add necessary import statements in colab notebooks

jinyongyoo:


Hanyu-Liu-123 (approved):
Looks good to me!

jxmorris12:
Hey @jinyongyoo -- shouldn't this happen already with `pip install textattack`? https://github.com/QData/TextAttack/blob/master/textattack/shared/utils/install.py#L117

PR #306, opened 2020-10-19
Update README -- how to customize augmenter

Hanyu-Liu-123:
One of our users asked in the question panel about how to create his own augmenter in python. This is actually very useful and pretty easy to do with just a few lines of code.


jxmorris12 (approved):
great example! And shows that an awesome API isn't useful for everyone without accompanying awesome documentation. Nice work!

Hanyu-Liu-123:
> great example! And shows that an awesome API isn't useful for everyone without accompanying awesome documentation. Nice work!

Thanks!

PR #318, opened 2020-10-27
add lazy_load to load Stanza

qiyanjun:
Due to the long loading time of the "import stanza", our readthedocs built failed due to long time/memory cost. So here is the lazy_load try to import stanza lazily. 

I have to merge this type of pull request more frequently because we set the readthedocs to build our documentation from the master branch. 
I don't know  other ways to know if a version I revised works well or not in readthedocs 
(my merged versions were all after extensive local built and tests)

### DIFF
--- a/examples/attack/attack_camembert.py
+++ b/examples/attack/attack_camembert.py
@@ -1,12 +1,12 @@
+# Quiet TensorFlow.
+import os
+
+import numpy as np
+from transformers import AutoTokenizer, TFAutoModelForSequenceClassification, pipeline
+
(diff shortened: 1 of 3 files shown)


jxmorris12:
It looks like the tests failed with an error:
```
ValueError: module object for 'stanza' substituted in sys.modules during a lazy load
```

qiyanjun:
readthedocs is so NICE!  They added us to a large build queue and triggered successful build for us. This has allowed for enough memory/time for our build. So no need to lazy-build stanza. Close.

jxmorris12:
Wow, that's awesome!!

PR #319, opened 2020-10-27
Transformation Fix: WordSwapNameChange

dheerajrav:
This fix accounts for sentences with just the firstname, or the lastname in them. 

`S-PER` NER was not accounted for in the transformation.

### DIFF
--- a/textattack/transformations/word_swap_change_name.py
+++ b/textattack/transformations/word_swap_change_name.py
@@ -52,13 +52,13 @@ class WordSwapChangeName(WordSwap):
         replacement_words = []
         tag = word_part_of_speech
         if (
-            tag.value == "B-PER"
+            tag.value in ("B-PER", "S-PER")
             and tag.score >= self.confidence_score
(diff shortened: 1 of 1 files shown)


jxmorris12:
Thanks @dheerajrav, great catch!

jxmorris12 (approved):

dheerajrav:
@jxmorris12 can we run the build tests again? Not sure why the test failed. they passed in my local machine.

jinyongyoo:
Hi @dheerajrav, the test that is failing actually has nothing to do with TextAttack. It's because Flair package that we use does not work with torch 1.7.0 for some reason (https://github.com/flairNLP/flair/issues/1924). Therefore, we temporarily set torch version to 1.6.0 in our requirements.txt. Please rebase this branch onto master branch and this should fix the issue.

dheerajrav:
@jinyongyoo done!

jinyongyoo:
Great, thank you!

PR #322, opened 2020-10-29
a comprehensive clean up of all URLs by using the :ref: / Plus Remove the old API docs

qiyanjun:
I have updated (almost) all code files' head docstring by adding text from the old API rst files. So we really don't need those files under 3import.

The only exception is for the attack recipe. So I keep that as a single item in the sidebar

Besides, I just did a comprehensive clean up of all URLs by using the :ref:


qiyanjun:
the error is from calling predict from the flair library in the CheckList augmenter test (not relevant to documentation )

qiyanjun:
There seems to be another snippet of code that uses flair to call predict in textattack/constraints/grammaticality/part_of_speech.py

qiyanjun:
Hanyu figured: 

seems from flair issues, switching to version 1.6 of torch seems to solve the problem.

https://github.com/flairNLP/flair/issues/1924

qiyanjun:
All tests passed. This merge also solves the current flair and flow inconsistency issues.

PR #324, opened 2020-11-04
custom dataset loader

ArshdeepSekhon:
loads a user specified dataset from local file as text attack dataset using hugging face data loader utils:
1. Adds a CustomDataset class to textattack datasets module
2. Converts a user specified dataset into a text attack dataset using hugging face data loader utils
3. Allows user to specify input and output columns from a local file/ pandas data frame/ dict
4. accordingly changes the augment command line from csv reader to CustomDataset


qiyanjun:
@ArshdeepSekhon is this a continued PR from 310? If yes, I will close that one. 

@ArshdeepSekhon there are some issues with the flair library dependency, so we fix that in the https://github.com/QData/TextAttack/pull/322

ArshdeepSekhon:
@qiyanjun This is a different PR , #310 adds an additional command to let the user specify `eval` on an entire dataset instead of sampling a subset.

textattack/datasets/custom_dataset.py
```diff
@@ -0,0 +1,166 @@
...
+    if not isinstance(s, str):
+        s = "custom " + str(type(s))
+    return textattack.shared.utils.color_text(str(s), color="blue", method="ansi")
+
+
+class CustomDataset(TextAttackDataset):
+    """Loads a Custom Dataset from a file/list of files and prepares it as a
+    TextAttack dataset.
+
+    - name(Union[str, dict, pd.DataFrame]): the user specified dataset file names, dicts or pandas dataframe
```
jinyongyoo, line 22: Don't you think `name` argument would be too confusing? Maybe `file_name_or_data`?

jinyongyoo (changes requested):
I didn't know `datasets` offered an easy way to load datasets from file, so this CustomDataset class was cool to see! I made some minor comments about handling some edge cases.

The biggest comment that I have is that after reading the code for this class, I'm still not completely sure how to use it. I think part of it can be resolved with more detailed documentation. But also, I think part of it is that we're abstracting too much away from the users. We are coupling `CustomDataset` so tightly with Huggingface's `dataset` package. Part of me feels like it would easier for users if we can just require users to create the appropriate `datasets.Dataset` object and pass it directly instead of requiring them to provide the arguments that are eventually used to build the `datasets.Dataset` object. After all, Huggingface has better documentation and examples for guiding users how to build the dataset. Then, we can simply skip most of the complicated logic in `__init__` and just focus on packing the examples as `OrderedDict` for TextAttack.

qiyanjun:
@ArshdeepSekhon It is a good idea to sync your copy of the code with the master repository regularly. This way you can quickly account for changes:

$ git remote add upstream https://github.com/QData/TextAttack
$ git fetch upstream
$ git rebase upstream/master

qiyanjun:
@ArshdeepSekhon please run

"black textattack" to correct all formatting issues first.
"isort textattack" to correct all import issues then.
"make test" then

qiyanjun:
@ArshdeepSekhon let me fix the build doc

qiyanjun:
@ArshdeepSekhon @jinyongyoo I thought datasets/dataset.py should be @abstractclass with some must @abstractmethod ..

```
import abc

class AbstractClass(metaclass=abc.ABCMeta):

    @abc.abstractmethod
    def abstractMethod(self):
        return
```

qiyanjun:
@ArshdeepSekhon @jinyongyoo I recommend we triangle this PR for now.. Agreed?

PR #326, opened 2020-11-06
added table with attack recipes

tahmid-kazi:


jxmorris12 (approved):
This is a good improvement!

@jinyongyoo do you think we should remove the bullet-point list of attack recipes, and just keep the table instead? Or should we include both?

jinyongyoo:
I think we can replace the bullet-point list with the table. But I'm not sure what the `Accessibility` column is about. It seems to be about the goal function we are maximizing, right? I'm not sure if that's particularly helpful, since all of them have the same value "score". Instead, I think we should try to put some other pieces of information, such as search method, blackbox vs whitebox, and maybe even transformation (though maybe that'll be confusing with the `Perturbation` column).

qiyanjun:
@tahmid-kazi can you update the table wrt @jinyongyoo comments?

qiyanjun:
I will merge this first, and we can add more columns later.

PR #329, opened 2020-11-07
turn unknown goal function error into a logger warning

a1noack:
I was trying to build an attack with a modified goal function, but I was unable to because of the errors raised in the `validate_model_goal_function_compatibility` function in textattack/shared/validators.py. I think these logger warnings will make it easier for people to experiment with custom goal functions.

### DIFF
--- a/textattack/goal_functions/goal_function.py
+++ b/textattack/goal_functions/goal_function.py
@@ -36,7 +36,7 @@ class GoalFunction(ABC):
         model_cache_size=2 ** 20,
     ):
         validators.validate_model_goal_function_compatibility(
-            self.__class__, model_wrapper.__class__
+            self.__class__, model_wrapper.model.__class__
         )
         self.model = model_wrapper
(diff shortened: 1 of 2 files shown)


uvafan (approved):
Good idea! We really shouldn't have been throwing an error in the first place in hindsight. This looks good to me, @jxmorris12 does it also look good to you?

jinyongyoo:
I also want to raise another issue that we can fix with this PR. The way we check compatibility for `seq2seq` is outdated/wrong since we actually don't have `textattack.models.translation` module or `textattack.models.summarization` modules anymore. 

If you run `textattack attack --goal-function targeted-classification --recipe textfooler --model t5-en-de`, you'll notice that the compatibility check doesn't work as intended.

jxmorris12 (approved):

jxmorris12 (approved):
these are small but very important improvements! thanks @a1noack and @jinyongyoo

PR #330, opened 2020-11-08
update package versions

jinyongyoo:
Previously, there was an issue with flair package and torch 1.7.0. Flair authors have fixed the issue, so newest flair should work with newest torch.


Hanyu-Liu-123 (commented):
Looks good to me! I'm glad flair resolved the issue

Hanyu-Liu-123 (approved):

Issue #325, opened 2020-11-05
Unexpected behaviour when using customized word embedding


tsinggggg:
I was using `TextFoolerJin2019` to attack the distillbert sentiment analysis model from huggingface transformer `    from transformers import pipeline
    p = pipeline("sentiment-analysis")`, but i downloaded `https://s3.amazonaws.com/dl4j-distribution/GoogleNews-vectors-negative300.bin.gz ` and replaced the original default paragram cf embedding in the recipe(one word embedding distance constrait and one word swap embedding transformation). I was using ['this is a bad movie', 0] as the input data.

The thing is in this vocab there are words like this: 'coulda_shoulda_woulda' and 'slovakia_slota_commented'. And among the attackedtexts, in some of them 'bad'  is replaced with complex words like these two. When this happens, these long words got tokenized and got broken into multiple words. One consequence is the 'newly_modified_indices' don't correspond correctly to the original sentence anymore. Therefore, list index of bound error occurs when checking the contraint.

To be specific, in this case where reference text is 'this is a bad movie' and attackedtext is 'this is a slovakia_slota_commented movie', in the code https://github.com/QData/TextAttack/blob/master/textattack/constraints/semantics/word_embedding_distance.py#L134, newly_modified_indices are 3,4,5 (representing slovakia, slota, commented after tokenization) and the reference_text has only length 5, therefore 5 will be out of bound.

I understand this maybe an unexpected way to use textattack, but i think it goes back to my previous wish list that textattack comes with its own ExceptionResult for any unexpected exceptions.

Also i think it's better to use defaultdict(dict) as the default for cos_sim_mat and mse_sim_mat, otherwise the statement in except block can fail:https://github.com/QData/TextAttack/blob/master/textattack/constraints/semantics/word_embedding_distance.py#L106
this test case fails
```
c = WordEmbeddingDistance(min_cos_sim=0.5)
c.cos_sim_mat = dict()
c.get_cos_sim(1, 10)
print(c.cos_sim_mat)
```
but this one passes
```
c = WordEmbeddingDistance(min_cos_sim=0.5)
c.cos_sim_mat = defaultdict(dict)
c.get_cos_sim(1, 10)
print(c.cos_sim_mat)
```

Thanks,

tsinggggg:
One potential solution for this particular issue may be simply returning false when there are more than 1 newly_modified_indice

[...]

PR #333, opened 2020-11-12
custom embeddings

tsinggggg:
1. extended the WordEmbedding class to implement 2 ways to initialize a word embedding: one is as before downloading from s3 all the files needed(embeddingmat, cos_sim_mat, etc..); the other being using gensim keyedvector to avoid re-implementing things(example in test case) 
2. incorporated WordEmbeddings with WordSwapEmbedding and WordEmbeddingDistance and ThoughtVector as per https://github.com/QData/TextAttack/issues/325#issuecomment-724888184. APIs of these 3 class are unchanged except an additional argument in __init__ to indicate whether to use embeddings from textattack s3 or in gensim keyedvectors format.

### DIFF
--- a/textattack/shared/word_embedding.py
+++ b/textattack/shared/word_embedding.py
@@ -3,12 +3,15 @@ Shared loads word embeddings and related distances
 =====================================================
 """
 
+from collections import defaultdict
 import os
 import pickle
 
 import numpy as np
+import torch
 
 import textattack
+from textattack.shared import utils
 
 
 class WordEmbedding:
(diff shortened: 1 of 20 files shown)


jinyongyoo:
@tsinggggg Thanks for the PR! Do you mind if I make some changes too? I like how you integrated gensim embeddings, but I also think there's a more general way we can set up the `WordEmbedding` class so it can work with other custom embeddings.

tsinggggg:
> @tsinggggg Thanks for the PR! Do you mind if I make some changes too? I like how you integrated gensim embeddings, but I also think there's a more general way we can set up the `WordEmbedding` class so it can work with other custom embeddings.

Not at all. I'd love to see a more general solution as well! @jinyongyoo

jinyongyoo:
@tsinggggg 

I felt that the API for creating `WordEmbedding` object was unclear, and it was even less clear if the user wanted to use their own word embedding that might be a different format. So I made `WordEmbedding` a base class so users know exactly what methods they to implement to use their own custom word embeddings. Also `TextAttackWordEmbedding` implements most of the necessary functions, so users can plug in appropriate components and have a word embedding class that works without implementing their own. Lastly `GensimWordEmbedding` provides a clear way to use gensim's `KeyedVectors`. Please let me know if you have any questions.

@jxmorris12 
Can you review this? I was planning to review it, but I ended up writing new stuff myself.

qiyanjun:
@jxmorris12 can you review?

textattack/shared/word_embedding.py
```diff
@@ -3,103 +3,437 @@
...
+            neighbours (list[int]): List of indices of the nearest neighbours
+        """
+        raise NotImplementedError()
+
+    __repr__ = __str__ = utils.default_class_repr
+
+
+class TextAttackWordEmbedding(WordEmbedding):
+    """Object for loading word embeddings and related distances for TextAttack.
+    This class has a lot of internal components (e.g. get consine similarity)
```
jxmorris12, line 105: typo: cosine

jxmorris12 (approved):
LGTM, but it's not immediately obvious what the difference between `WordEmbedding` and  `TextAttackWordEmbedding` is, or why they're separate classes in the first place.

jinyongyoo:
> LGTM, but it's not immediately obvious what the difference between `WordEmbedding` and `TextAttackWordEmbedding` is, or why they're separate classes in the first place.

Well `WordEmbedding` defines what methods need to implemented if a user wants to come up with a custom word embedding class. `TextAttackWordEmbedding` is the class that actually has a lot of the internal stuff implemented and just requires the user to pass in the appropriate matrices and dictionary. Most users should just use `TextAttackWordEmbedding`, but `WordEmbedding` is there just in case some really wants to implement their own methods.

jxmorris12:
> > LGTM, but it's not immediately obvious what the difference between `WordEmbedding` and `TextAttackWordEmbedding` is, or why they're separate classes in the first place.
> 
> Well `WordEmbedding` defines what methods need to implemented if a user wants to come up with a custom word embedding class. `TextAttackWordEmbedding` is the class that actually has a lot of the internal stuff implemented and just requires the user to pass in the appropriate matrices and dictionary. Most users should just use `TextAttackWordEmbedding`, but `WordEmbedding` is there just in case some really wants to implement their own methods.

I see. That makes sense. It seems a little arbitrary which one is which name-wise, is all.

PR #334, opened 2020-11-12
fix inflection generation and enlarge pos to upos mapping (for Morpheus)

a1noack:
I think `textattack/transformations/word_swap_inflections.py` had a couple issues. [These are issues in the sense that the Morpheus attack&ndash;that relies on this transformation&ndash;is not a correct copy of the authors' original implementation of Morpheus when it uses this transformation.]
Specifically:
**1.** The function `_get_replacement_words` wasn't returning the correct lists of inflections. For example, notice that
```
>>> lemminflect.getAllLemmas('is')
{'AUX': ('be',), 'VERB': ('be',)}
```

However, according to the Morpheus paper, we want not only 'be', but 'been', 'am', 'was', 'are', 'were', and 'being'. To get this full list, you must take one of the lemmas returned by `lemminflect.getAllLemmas` (in this case, 'be') and get its inflections: 

```
>>> lemminflect.getAllInflections('be', upos='VERB').values()
dict_values([('be',), ('was', 'were'), ('being',), ('been',), ('am', 'are'), ('is',)])
```
**2.** The number of considered POS tags was too low. The result was that many words that had inflections were being skipped over and not considered for replacement. Therefore, I expanded the mapping using the en-ptb (English Penn Treebank) POS to universal POS mapping introduced [here](https://github.com/slavpetrov/universal-pos-tags/blob/master/en-ptb.map).

### DIFF
--- a/textattack/transformations/word_swap_inflections.py
+++ b/textattack/transformations/word_swap_inflections.py
@@ -29,16 +29,16 @@ class WordSwapInflections(WordSwap):
         super().__init__(**kwargs)
         # fine-grained en-ptb POS to universal POS mapping
         # (mapping info: https://github.com/slavpetrov/universal-pos-tags)
-        self._enptb_to_universal = {'JJRJR': 'ADJ', 'VBN': 'VERB', 'NN': 'NOUN', 'VBG': 'NOUN',
-                                  'VBP': 'VERB', 'JJ': 'ADJ', 'VBZ': 'VERB',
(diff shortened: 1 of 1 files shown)


jxmorris12 (approved):
Thanks for the detailed explanation, and commenting your code! This looks good to me!

jxmorris12:
@a1noack run `make format` to format your code please. Then we can merge!

a1noack:
Oops I had the wrong isort.

jinyongyoo:
@a1noack  Thanks for fixing the issue!

PR #336, opened 2020-11-13
Fix ReadMe Minor Mistakes

Hanyu-Liu-123:
There's a typo I made with the augmenter example.

Also we now have five augmentation recipes instead of three!


jinyongyoo (approved):

jxmorris12:
Thank you!

PR #338, opened 2020-11-14
update the readme table of recipes and try to fix the readthedoc build error

qiyanjun:


Opdoop:
What's the reason to change readme table from markdown table to the HTML table?

Opdoop:
Assume try to use html style make the table more readable. 
But, as describle in [stackflow](https://stackoverflow.com/questions/51956361/custom-css-file-for-readme-md-in-a-github-repo) and  [github markup](https://github.com/github/markup) , for security reason, the github markdown file will be `sanitized, aggressively removing things that could harm you and your kin—such as script tags, inline-styles, and class or id attributes.`

qiyanjun:
Good point! I took a quick and easier way to get the table look well .. yes. Pure markdown is better. Will do later!

Issue #341, opened 2020-11-17
pip install bug


Opdoop:
### Bug 
install textattack through pip missing dependent package
### Enviroment
Ubuntu 18.04.5 LTS
conda 4.8.3
### Reproducing
```
conda create -n textattack python=3.6
conda activate textattack
pip install textattack
textattack --help
```
Then encounter this:
```
textattack: Updating TextAttack package dependencies.
textattack: Downloading NLTK required packages.
[nltk_data] Downloading package averaged_perceptron_tagger to
[nltk_data]     /home/nano/nltk_data...
[nltk_data]   Unzipping taggers/averaged_perceptron_tagger.zip.
[nltk_data] Downloading package stopwords to /home/nano/nltk_data...
[nltk_data]   Unzipping corpora/stopwords.zip.
[nltk_data] Downloading package omw to /home/nano/nltk_data...
[nltk_data]   Unzipping corpora/omw.zip.
[nltk_data] Downloading package universal_tagset to
[nltk_data]     /home/nano/nltk_data...
[nltk_data]   Unzipping taggers/universal_tagset.zip.
[nltk_data] Downloading package wordnet to /home/nano/nltk_data...
[nltk_data]   Unzipping corpora/wordnet.zip.
[nltk_data] Downloading package punkt to /home/nano/nltk_data...
[nltk_data]   Package punkt is already up-to-date!
Traceback (most recent call last):
  File "/home/nano/anaconda3/envs/textattack/bin/textattack", line 5, in <module>
    from textattack.commands.textattack_cli import main
  File "/home/nano/anaconda3/envs/textattack/lib/python3.6/site-packages/textattack/__init__.py", line 12, in <module>
    from . import (
  File "/home/nano/anaconda3/envs/textattack/lib/python3.6/site-packages/textattack/attack_recipes/__init__.py", line 23, in <module>
    from .bae_garg_2019 import BAEGarg2019
  File "/home/nano/anaconda3/envs/textattack/lib/python3.6/site-packages/textattack/attack_recipes/bae_garg_2019.py", line 6, in <module>
    from textattack.constraints.grammaticality import PartOfSpeech
  File "/home/nano/anaconda3/envs/textattack/lib/python3.6/site-packages/textattack/constraints/__init__.py", line 24, in <module>
    from . import grammaticality
  File "/home/nano/anaconda3/envs/textattack/lib/python3.6/site-packages/textattack/constraints/grammaticality/__init__.py", line 10, in <module>
[...]

PR #342, opened 2020-11-17
Update requirements.txt

Opdoop:
Fix the bug `No module named 'google'` found in issue https://github.com/QData/TextAttack/issues/341

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,3 +19,4 @@ word2number
 num2words
 more-itertools
 PySocks!=1.5.7,>=1.5.6
+google-api-python-client


Opdoop:
Fix the bug found in issue https://github.com/QData/TextAttack/issues/341

qiyanjun:
@jinyongyoo can you double-check on this? I met the same error in a fresh installation and used the same trick solved. But unsure if that is needed in the requirement.txt since you have updated the lazy-loader to fix that?

qiyanjun:
@Opdoop can you make a fresh installation? We actually tried to fix this by a lazy_loading style. That seems to work well.

jinyongyoo:
@Opdoop Thanks for raising this issue. We actually fixed this issue in PR #339 on Saturday, one day we released new version on PyPI. We'll release a new version today.

jinyongyoo:
Released version 0.2.13 on PyPI.

PR #349, opened 2020-11-19
update pos tagging in AttackedText

jinyongyoo:


textattack/shared/utils/strings.py
```diff
@@ -194,7 +194,7 @@ def zip_flair_result(pred, tag_type="pos-fast"):
     pos_list = []
     for token in tokens:
         word_list.append(token.text)
-        if tag_type == "pos-fast":
+        if "pos" in tag_type:
             pos_list.append(token.annotation_layers["pos"][0]._value)
```
Hanyu-Liu-123, line 198: Do we need to change this line? I'm not sure what's the difference between `upos-fast` and `pos-fast`...

Hanyu-Liu-123 (approved):
What's the difference between `pos-fast` tagger and `upos-fast` tagger? Otherwise, looks good to me :)

textattack/shared/utils/strings.py
jinyongyoo: I changed it to `"pos" in tag_type` since it's more general than check if it is `pos-fast`. Yes we still retrieve the POS value using "pos".

Hanyu-Liu-123: Ok Thanks!

jinyongyoo:
@Hanyu-Liu-123  The main difference are the tagsets. `upos-fast uses Universal tagset, while `pos-fast` users Brown Corpus (?) tagset.  Different tagsets mean that the values representing different part of speech are different. For example, adjective is represented by `ADJ` in universal tag set and by `JJ` in Brown Corpus tagset.

PR #351, opened 2020-11-20
Updated model zoo tables

tahmid-kazi:
Updated model zoo table with a more organized structure for all the relevent information


qiyanjun:
this does not add more information. therefore closed

PR #356, opened 2020-11-23
Add CLARE Attack

jinyongyoo:
This PR adds [CLARE](https://arxiv.org/abs/2009.07502) attack, which uses distilled RoBERTa masked language model to perform word swaps, word insertions, word merges (which is where we combine two adjacent words and replace it with another word) in a greedily manner. 

New additions:
- `clare` recipe
- `WordInsertionMaskedLM`, which is similar to `WordSwapMaskedLM` but instead insert new words.
- `WordMergeMasedLM`
-  Base classes `WordInsertion` and `WordMerge`.

Changes:
- `WordSwapMaskedLM` has been updated to have a minimum confidence score cutoff and batching has been added for faster performance.
- Changes in directory structure for transformations.


qiyanjun:
@jinyongyoo flake8 textattack/
textattack/transformations/__init__.py:10:1: F403 'from .word_swaps import *' used; unable to detect undefined names
textattack/transformations/__init__.py:11:1: F403 'from .word_insertions import *' used; unable to detect undefined names
textattack/transformations/__init__.py:12:1: F403 'from .word_merges import *' used; unable to detect undefined names

qiyanjun:
@Hanyu-Liu-123 please review!

Hanyu-Liu-123:
> @Hanyu-Liu-123 please review!

will do today!

textattack/transformations/word_merges/word_merge_masked_lm.py
```diff
@@ -0,0 +1,184 @@
...
+        indices_to_modify = list(indices_to_modify)
+        # find indices that are suitable to merge
+        token_tags = [
+            current_text.pos_of_word_index(i) for i in range(current_text.num_words)
+        ]
+        merge_indices = find_merge_index(token_tags)
+        merged_words = self._get_merged_words(current_text, merge_indices)
+        transformed_texts = []
+        for i in range(len(merged_words)):
+            index_to_modify = indices_to_modify[i]
```
Hanyu-Liu-123, line 145: I think this line should actually be `index_to_modify = merge_indices[i]`. We're trying to merge the indices specified by `merge_indices` and not `indices_to_modify`. Those 2 lists don't necessarily match.

for example, suppose we have the sentence **"I love to play soccer"**, `merge_indices = [3]`, `indices_to_modify = [0,1,2,3,4]`, and `merged_words = ["sleep"]`. Calling `index_to_modify = indices_to_modify[i]` would give us **"sleep to play soccer"**, which is wrong. The correct output should be **"I love to sleep"**.

Hanyu-Liu-123 (changes requested):
Looks good to me except for that one line! 

Also, was the weird character problem solved by just stripping away those characters?

Thanks!

textattack/transformations/word_merges/word_merge_masked_lm.py
cookielee77: I agree. 

Also, does merge_indices have an order? e.g., which index should be merged first? some indices should have higher priority since they are easily attacked. 

Additionally, could you let me know which line of the code has the function that orders the three perturbations at the same indice? e.g., at [3], maybe the perturbation effect has: Insert > Replace > Merge. In this case, we only use Insert to perturb the text and will discard the other two perturbations.

textattack/transformations/word_merges/word_merge_masked_lm.py
```diff
@@ -0,0 +1,184 @@
...
+                        temp_text.replace_word_at_index(index_to_modify, word)
+                    )
+
+        return transformed_texts
+
+    def extra_repr_keys(self):
+        return ["masked_lm_name", "max_length", "max_candidates", "min_confidence"]
+
+
+def find_merge_index(token_tags, indices=None):
```
cookielee77, line 163: This was my developed code. You can try to relax the constraints or discard the function to see whether the generated texts look reasonable.

textattack/attack_recipes/clare_li_2020.py
```diff
@@ -0,0 +1,108 @@
...
+        #
+        # Don't modify the same word twice or stopwords.
+        #
+        constraints = [RepeatModification(), StopwordModification()]
+
+        # "A  common  choice  of sim(·,·) is to encode sentences using neural networks,
+        # and calculate their cosine similarity in the embedding space (Jin et al., 2020)."
+        # The original implementation uses similarity of 0.7. Since the CLARE code is based on the TextFooler code,
+        # we need to adjust the threshold to account for the missing / pi in the cosine similarity comparison
+        #  So the final threshold is 1 - (1 - 0.7) / pi =  0.904507034.
```
cookielee77, line 83: So my original threshold doesn't have the PI in cosine similarity. Not sure whether it has some significant impacts.

cookielee77 (commented):
I'm not familiar with the whole codebase and pipeline. So I just leave some high-level comments when compared with my original implementations.

textattack/transformations/word_merges/word_merge_masked_lm.py
Hanyu-Liu-123: > I agree.
> 
> Also, does merge_indices have an order? e.g., which index should be merged first? some indices should have higher priority since they are easily attacked.
> 
> Additionally, could you let me know which line of the code has the function that orders the three perturbations at the same indice? e.g., at [3], maybe the perturbation effect has: Insert > Replace > Merge. In this case, we only use Insert to perturb the text and will discard the other two perturbations.

`merge_indices` does not have a priority order, since it is just the output of `find_merge_index`. `find_merge_index` runs on the modifiable indices of the text (something like [0,1,2,5]) and is in ascending order of numerical value.

We actually generate all merged sentences first, and then feed them along with all the perturbed sentences generated from `masked_insertion` and `masked_swap` to the search function. So we don't really order the three perturbations and just generate every eligible perturbation, then find the best one using the search function.

textattack/attack_recipes/clare_li_2020.py
```diff
@@ -0,0 +1,108 @@
...
+        shared_tokenizer = transformers.AutoTokenizer.from_pretrained(
+            "distilroberta-base"
+        )
+        transformation = CompositeTransformation(
+            [
+                WordSwapMaskedLM(
+                    method="bae",
+                    masked_language_model=shared_masked_lm,
+                    tokenizer=shared_tokenizer,
+                    max_candidates=20,
```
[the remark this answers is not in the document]
jinyongyoo: `max_candidates` is like taking top-k after filtering by min confidence. I set it to 50.

textattack/attack_recipes/clare_li_2020.py
jinyongyoo: I think the issue is whether you used cosine similarity or angular similarity. Did you use cosine similarity? Then it should just be 0.7.

cookielee77: Yes, I just used cosine similarity

Hanyu-Liu-123 (approved):

PR #361, opened 2020-12-03
Change check_robustness dataset

Opdoop:
As Adversarial training using training data, although here has a random sample, it's more reasonable to use eval data to check robustness.


qiyanjun:
@Opdoop let me check the doc error.

qiyanjun:
@Opdoop please rebase

Opdoop:
@qiyanjun Rebased

jxmorris12 (approved):
I think that evaluating adversarial robustness using the evaluation dataset is a reasonable default!

PR #364, opened 2020-12-05
Update make-docs.yml

qiyanjun:
from: 

 pip install -e . ".[dev]" # This should install all packages for development

to 
        pip install -e .[dev]


qiyanjun:
@jxmorris12 in the past week, somehow the build-doc check always gave install dependency errors, though every doc builds well locally. 

I made the above change in build-doc workflow. Is this change correct?

qiyanjun:
@jxmorris12 @jinyongyoo I mess up the git-workflow by trying the changes in the master branch already. ah. certainly need to be more mindful in following PR workflow.

PR #368, opened 2020-12-10
Fix RandomSwap and RandomSynonymInsertion bug

DerekChia:
**Problem**
When using `EasyDataAugmenter`, there are two cases that will cause the augmenter to run into an infinite loop. 

Case 1: Text that contains duplicated words e.g. `hello hello`

For a sentence / phrase containing duplicated words (`hello hello`), `RandomSwap` would not be able to successfully perform the swap since both words are the same. This causes `words_swapped` to never reach 0 and hence the while loop will never reach its break condition. 

Case 2: Text that contains words with no synonym e.g. `hey derek`

For a sentence / phrase e.g. `hey derek` where there is no synonym that can be inserted, `RandomSynonymInsertion` would not be able to successfully insert a synonym even after 7 attempts. Similarly, this means that `words_swapped` would never reach 0 and hence the while loop will never reach its break condition. 

Here is the while loop which causes the problem:

https://github.com/QData/TextAttack/blob/0cc049a8c77f7f17c9642ac09ca1b60aa17645e8/textattack/augmentation/augmenter.py#L84-L106

**Reproducing the Problem**
Running the following will cause TextAttack augment to enter an infinite loop.

```
from textattack.augmentation import EasyDataAugmenter
augmenter = EasyDataAugmenter()
s = `'hello hello' # case 1
s = 'hey derek' # case 2
l = augmenter.augment(s)
print(l)
```

**Fix**

Case 1

We add `len(set(transformed_texts)) == 1` to check if the entire string contains duplicated word. If so, break out of the while loop. 

Case 2

We also break out of the while loop when `transformed_texts == [current_text]`. This happens when the `transformed_texts` is same as the original text (`current_text`) returned after 7 attempts.

https://github.com/QData/TextAttack/blob/0cc049a8c77f7f17c9642ac09ca1b60aa17645e8/textattack/transformations/random_synonym_insertion.py#L31-L36

**Result**
With the fix, running TextAttack augment with `EasyDataAugmenter` will output the expected result.

Case 1
```
['hello hello', 'hello hello howdy', 'hello', 'hello hi']
```

Case 2
```
['derek', 'hey derek', 'derek hey']
```

p.s. also added `augmented_text = list(set(augmented_text))` to remove duplicated results in `EasyDataAugmenter`.


qiyanjun:
@DerekChia please rebase.. and fix the errors...

qiyanjun:
@DerekChia do you know why some test failed?

Hanyu-Liu-123:
> @DerekChia do you know why some test failed?

I think it is because those 2 fixes will terminate the while loop earlier than we want it to, **when we're not using the `EasyDataAugmenter`.**

For example, for other augmenters, having duplicated words like **"hello hello"** will not cause infinite loop so the augmenting process should not be terminated.

What I think that can fix the infinite loop issue would be to increase `words_swapped` by at least 1 during each iteration of the while loop. We can replace this [line](https://github.com/QData/TextAttack/blob/994a56caf29c7e0162d8fa5ec5d58eab58d30daa/textattack/augmentation/augmenter.py#L106) (`words_swapped = len(current_text.attack_attrs["modified_indices"])`) with `words_swapped = max(len(current_text.attack_attrs["modified_indices"]), words_swapped+1)`. 

In this way, the while loop will iterate at most `num_words_to_swap` times (num_words_to_swap < length of the sentence). It is not pretty but should be able to solve the infinite loop issue without affecting other augmenters.

qiyanjun:
@DerekChia mind to answer @Hanyu-Liu-123 question?

DerekChia:
Let me look into it. Thanks for highlighting.

qiyanjun:
@DerekChia possible to confirm?

qiyanjun:
@DerekChia please rebase

qiyanjun:
@Hanyu-Liu-123 is this still concerning? I believe we fixed the issue in the recent PR?

Hanyu-Liu-123:
> @Hanyu-Liu-123 is this still concerning? I believe we fixed the issue in the recent PR?

I tried running DerekChia's tests above and the error is still reproduced. Which PR was this issue addressed?

qiyanjun:
This is solved by a new PR #477

qiyanjun:
This is solved by a new PR #477

PR #392, opened 2020-12-17
Clare augmentation recipe

Hanyu-Liu-123:
This is the augmentation recipe from CLARE, which uses the following transformations:

- WordSwapMaskedLM
- WordInsertionMaskedLM
- WordMergeMaskedLM

The default model and tokenizer are both `distilroberta-base`. Specific models/tokenizers can be specified by the user if they decide to call `CLAREAugmenter` in python.


jinyongyoo (approved):
LGTM!

PR #409, opened 2021-01-06
Fix problem with file redirection (option to disable color)

cogeid:
# What does this PR do?

## Summary
Adds an option to disable color when printing results by typing --nocolor in the command line.

## Additions
Added a global variable color in attack_result.py, when the variable is set to false it disables colors when printing results
Added a new arg in attack_command.py. The arg is called "nocolor", which has action = store_true
Added a new statement in textattack_cli.py, determines if the user typed the argument --nocolor, and sets the global variable color as true if the user has entered --nocolor, else false

## Changes
None

## Deletions
None

Relevant Tests still needed


qiyanjun:
@cogeid thanks.. Please rebase.. and also correct the pytest errors

jinyongyoo (changes requested):
@cogeid Thanks for trying to address the issue. Unfortunately, this is probably not the best way to add `nocolor` option. 

The module that actually prints the `AttackResult` to terminal is done by `textattack.loggers.FileLogger`. If you read its `log_attack_result` function, you'll see how color method is set for stdout. The best way to implement no color option would be the modify the `FileLogger` instead of the `AttackResult` class. If you modify `AttackResult`, there will be no differentiation for other logs (e.g. CSV, text logs). Color method isn't just for coloring text for terminal output. It also determines how we mark the differences in original text and perturbed text for files. For terminal outputs, we still probably will want to mark the differences, so one option would be to use `"file"` as `color_method`. Also, it's much better to work with class attributes instead of setting a global variable.

Lastly, there is a smarter way to enforce nocolor option when redirecting stdout. `sys.stdout.isatty()` lets you know if your program is actually printing to the terminal or to a different file. If `sys.stdout.isatty()` is false, we can go ahead and disable coloring across TextAttack (even loggers). 

We definitely do want to add this as an option though, so if you address the issues, I'll be more than happy to take another look.

textattack/loggers/file_logger.py
```diff
@@ -14,6 +14,8 @@
 class FileLogger(Logger):
     """Logs the results of an attack to a file, or `stdout`."""
 
+    nocolor = False
```
jinyongyoo, line 17: Can you make `nocolor` an attribute of the specific `FileLogger` instance (e.g. `self.nocolor`) instead of the entire class? Remember that if you set the `nocolor` a class-level attribute, then it's going to set `color_method` to `None` for actual text file logging too.

textattack/commands/textattack_cli.py
```diff
@@ -47,6 +49,12 @@ def main():
 
     # Let's go
     args = parser.parse_args()
+    if "--nocolor" in sys.argv:
```
jinyongyoo, line 52: Can you handle this in `parse_logger_from_args` method?

textattack/loggers/file_logger.py
```diff
@@ -43,6 +45,8 @@ def __setstate__(self, state):
     def log_attack_result(self, result):
         self.num_results += 1
         color_method = "ansi" if self.stdout else "file"
+        if self.nocolor:
+            color_method = None
```
jinyongyoo, line 49: I think it's better to set `color_method` to be `"file"` instead of `None`. `None` leaves no marks as to which words are different, but `"file"` marks it using double brackets ("[[ <word> ]]"). 

Also the logic needs to be clearer. Nocolor option should be check if `self.stdout` is true. Ex:
```
if self.stdout:
    color_method = "file" if self.nocolor else "ansi"
else:
    color_method = "file"
```

filetest.txt
```diff
@@ -0,0 +1,125 @@
+Attack(
```
jinyongyoo, line 1: I don't think you're actually running a test on this file. You need to update the pytest files.

textattack/loggers/file_logger.py
jinyongyoo: Could you also add the part the automatically enforces no color option when redirecting stdout? You can check via `sys.stdout.isatty()`. 

Also, for redirection case, could you also update this part https://github.com/QData/TextAttack/blob/994a56caf29c7e0162d8fa5ec5d58eab58d30daa/textattack/shared/utils/install.py#L95 so that there's no coloring around "textattack" prefix when we're redirecting?

jinyongyoo (changes requested):

textattack/loggers/attack_log_manager.py
```diff
@@ -27,7 +35,13 @@ def enable_wandb(self):
         self.loggers.append(WeightsAndBiasesLogger())
 
     def add_output_file(self, filename):
```
jinyongyoo, line 37: We don't need to specify color for file logging. I think the old one line code is fine.

textattack/loggers/attack_log_manager.py
```diff
@@ -18,7 +20,13 @@ def __init__(self):
         self.results = []
 
     def enable_stdout(self):
```
jinyongyoo, line 22: It's not desirable to directly access `nocolor` via `sys.argv`. We should not assume that `AttackLogManager` is necessary created using the command line.  All the command line argument parsing is done using the `argparse` library in `TextAttackCommand` modules which handle the command line executions.

jinyongyoo (changes requested):
This PR is very close

qiyanjun:
@cogeid can you update regarding Yoo’s comment?

cogeid:
> @cogeid can you update regarding Yoo’s comment?

Yes I have been working on it, I had some internet issues in my go local college dorm the past week.

qiyanjun:
@cogeid has the update been committed?

qiyanjun:
@cogeid please rebase

qiyanjun:
@cogeid please rebase your PR branch to the master

cogeid:
> @cogeid please rebase your PR branch to the master

I think I rebased it yesterday? Was there an error?

PR #406, opened 2020-12-31
Attack via API

jinyongyoo:
# What does this PR do?

## Summary
Performing attacks using the TextAttack API is quite limiting compared to using the command line commands because the users have to write their own code to support logging, saving checkpoint, and parallel processing. This PR intends to introduce some new support for attacking using TextAttack's API so that what can be done in command line can also be done using the API. However, this version does break a lot of backward compatibilities. 

## Additions
- Added new dataclasses ``AttackArgs``, ``AugmenterArgs``, ``DatasetArgs``, ``ModelArgs`` that are intended to represent arguments for specific features of TextAttack. 
- Added new module named `textattack.Attacker`.``Attacker`` takes in ``Attack``, ``textattack.datasets.Dataset``, and ``AttackArgs`` objects to perform attacks on the datasets with the same features for logging and saving checkpoints as the command line version.
- Added ``textattack.datasets.Dataset`` class. This takes in a list (or list-like) of tuples of ``(input, output)`` format. It is intended to the be the most basic class that other dataset classes extend from. The idea is to require all "dataset" inputs to be of type ``textattack.datasets.Dataset``, so it's clear the users what they need to pass as a "dataset" (behavior, an arbitrary list counted as a "dataset" technically). 

## Changes
- Removed ``attack_dataset`` method from ``Attack`` class and removed all the generator-related functionalities. Generator design is neat, but is less readable and maintainable. The idea is to remove all dataset-level attack features from ``Attack`` class and make ``Attacker`` handle the dataset level attacks. ``Attack`` is solely responsible for attacking single examples.
- Updated ``textattack.datasets.HuggingFaceDatasets`` to allow users to pass in their own ``datasets.Dataset`` objects (instead of just taking in ``name``, ``subset``, and ``split`` to load the dataset).
- Updated modules in `textattack.commands` to work with the new APIs.

PR #428, opened 2021-03-07
API Rework

jinyongyoo:
# What does this PR do?

Slides that cover the changes more in depth: https://docs.google.com/presentation/d/1n80DtBt-B77H9W39SngHJKQpuImrSF-DBLzJuFIS8Ao/edit?usp=sharing

## Summary
Current, TextAttack is primarily designed to work via command line. There is API-level attack functionality provided by `textattack.shared.Attack` class, but this only enables attacks individual examples and does not support attacking an entire dataset with logging, checkpoints, and parallelism supported.

Over time, it's becoming clearer than there is quite a gap between the features we provide via command line and via API.  However, to use TextAttack to its full potential, we need to better support API-level use of TextAttack. For example, it's very difficult to load custom attacks, models, and datasets via command line.

This big PR is intended to be a complete overhaul of TextAttack's API that improves user experience significantly aat the cost of breaking a lot of backward compatibility.

## Additions
- Added new dataclasses ``AttackArgs``, ``AugmenterArgs``, ``DatasetArgs``, ``ModelArgs`` that are intended to represent arguments for specific features of TextAttack. 
- Added new module named `textattack.Attacker`.``Attacker`` takes in ``Attack``, ``textattack.datasets.Dataset``, and ``AttackArgs`` objects to perform attacks on the datasets with the same features for logging and saving checkpoints as the command line version.
- Added ``textattack.datasets.Dataset`` class. This takes in a list (or list-like) of tuples of ``(input, output)`` format. It is intended to the be the most basic class that other dataset classes extend from. The idea is to require all "dataset" inputs to be of type ``textattack.datasets.Dataset``, so it's clear the users what they need to pass as a "dataset" (behavior, an arbitrary list counted as a "dataset" technically). 

## Changes
- Removed ``attack_dataset`` method from ``Attack`` class and removed all the generator-related functionalities. Generator design is neat, but is less readable and maintainable. The idea is to remove all dataset-level attack features from ``Attack`` class and make ``Attacker`` handle the dataset level attacks. ``Attack`` is solely responsible for attacking single examples.
- Updated ``textattack.datasets.HuggingFaceDatasets`` to allow users to pass in their own ``datasets.Dataset`` objects (instead of just taking in ``name``, ``subset``, and ``split`` to load the dataset).
- Updated modules in `textattack.commands` to work with the new APIs.


jinyongyoo:
Will's Update: https://github.com/QData/TextAttack/pull/406#issuecomment-765490831

### Notebook Examples that Pass vs. Fail - from latest commits
**Pass:**
- **Example 2:** Constraints
- **Example 4:** Tensorflow
- **Example 6:** AllenNLP
- **Example 7:** Multilingual
- **Example 8:** Attacking BERT models w/ CAptum



**Fail or Need Updating:**
- **Example 0**: End_To_End Tutorial
  - Issue with **peek-dataset** command: Missing args.model parameter
    - Possible Fix: Change parameter to dataset_by_model, include in parsed arguments
  - Issue with **eval** command: "PyTorch model must be torch.nn.Module, got type {type(model)}"
    - The model is expecting a torch.nn.Module, but instead the HuggingFaceModelWrapper is being passed
- **Example 1**: Introduction_and_Transformations
  - Issue with Attacking Custom Datasets, getting IndexOutOfBounds Error
- **Example 3:** Augmentation
- **Example 5:** Attacking Sklearn models - running out of RAM, try outside of Google Colab, if possible.
- **Example 9:** Keras example needs updating to be compatible with this branch

jinyongyoo:
Cogeid's Update: https://github.com/QData/TextAttack/pull/406#issuecomment-782191329

Notebook Tutorial

- Tutorial 1: Introduction_and_Transformations
https://textattack.readthedocs.io/en/latest/2notebook/1_Introduction_and_Transformations.html

Since the structures of some files and classes were changed in this PR, the original Tutorial has to be modified in order to run properly.

**The file attack.py has been moved from textattack.shared folder to textattack folder, so the original notebook code**

`from textattack.shared import Attack` **has to be changed to** `from textattack import Attack`

**The following lines must be added so that new data classes are imported**

`from textattack import AttackArgs`
`from textattack.datasets import Dataset`
`from textattack import Attacker`

the following lines were deleted

logger = CSVLogger(color_method='html')

```
num_successes = 0
while num_successes < 10:
    result = next(results_iterable)
    if isinstance(result, SuccessfulAttackResult):
        logger.log_attack_result(result)
        num_successes += 1
        print(f'{num_successes} of 10 successes complete.')
```

the lines above were replaced by

```
attack_args = AttackArgs(num_examples=10, checkpoint_interval=10, checkpoint_dir="checkpoints")

attacker = Attacker(attack, dataset, attack_args)

attacker.attack_dataset()
```
**Issue with "Bonus: Attacking Custom Samples"**
AttributeError: 'Dataset' object has no attribute 'output_scale_factor'
Dataset is lacking the output_scale_factor field

(Apparently this issue was fixed and merged after the meeting, I will be checking if the tutorial works after the fix)

jinyongyoo:
Tahmid's Comment: https://github.com/QData/TextAttack/pull/406#issuecomment-786578839

Notebook Tutorial (comments):

Notebook 5: Training the sklearn models fills up the RAM on Google Colab, so that part of the notebook needs to be updated to reduce space complexity. Model training is successful, but unable to test the Attack aspect (last cell) of the notebooks due to failure of the previous step.

Notebook 9: Attacking Keras models
https://textattack.readthedocs.io/en/latest/2notebook/Example_6_Keras.html
An error in the notebook occurs when setting up the Attack on the Keras model. Specifically, the error stems from the line:
`text_array = np.array([words2tokens(text_input) for text_input in text_input_list])`
where the error received is the following:
`NameError: name 'words2tokens' is not defined`
Model training is successful, but unable to test the Attack aspect (last 2 cells) of the notebooks due to failure of the previous step.

Hanyu-Liu-123:
> Will's Update: [#406 (comment)](https://github.com/QData/TextAttack/pull/406#issuecomment-765490831)
> 
> ### Notebook Examples that Pass vs. Fail - from latest commits
> **Pass:**
> 
> * **Example 2:** Constraints
> * **Example 4:** Tensorflow
> * **Example 6:** AllenNLP
> * **Example 7:** Multilingual
> * **Example 8:** Attacking BERT models w/ CAptum
> 
> **Fail or Need Updating:**
> 
> * **Example 0**: End_To_End Tutorial
>   
>   * Issue with **peek-dataset** command: Missing args.model parameter
>     
>     * Possible Fix: Change parameter to dataset_by_model, include in parsed arguments
>   * Issue with **eval** command: "PyTorch model must be torch.nn.Module, got type {type(model)}"
>     
>     * The model is expecting a torch.nn.Module, but instead the HuggingFaceModelWrapper is being passed
> * **Example 1**: Introduction_and_Transformations
>   
>   * Issue with Attacking Custom Datasets, getting IndexOutOfBounds Error
> * **Example 3:** Augmentation
> * **Example 5:** Attacking Sklearn models - running out of RAM, try outside of Google Colab, if possible.
> * **Example 9:** Keras example needs updating to be compatible with this branch

Turned out that the **Example 3 Augmentation** notebook failed because of a new update of **torch** [(link)](https://github.com/flairNLP/flair/issues/2137) because the `word_swap_change_location.py` uses **flair** (which uses **torch** and gives the error). When running the notebook, installing an older version of torch will solve the problem.

cogeid:
Notebook9 Update:

Imported Attacker, Dataset and AttackArgs to be compatible with the API rework.

Added following lines in the last cell for for compatibility reasons:

```
attack_args = AttackArgs(num_examples=10, checkpoint_dir="checkpoints")

attacker = Attacker(attack, dataset, attack_args)

attacker.attack_dataset()
```

> NameError: name 'words2tokens' is not defined

This issue was not due to the API rework, rather the implementation of call function. The function was using an undefined name, which I could not find the origin of, so I had to edit the call function so that text input is trained by the keras model and returns a prediction that is accepted by textattack.

    def __call__(self, text_input_list):
      
```
      x_transform = []
      for i, review in enumerate(text_input_list):
        tokens = [x.strip(",") for x in review.split()]
        BoW_array = np.zeros((NUM_WORDS,))
        for word in tokens:
          if word in vocabulary:
            if vocabulary[word] < len(BoW_array):
              BoW_array[vocabulary[word]] += 1            
        x_transform.append(BoW_array)
      x_transform = np.array(x_transform)
      prediction = self.model.predict(x_transform)
      return prediction
```

qiyanjun:
@jinyongyoo status? ready?

requirements.txt
```diff
@@ -1,6 +1,6 @@
 bert-score>=0.3.5
 editdistance
-flair==0.6.1.post1
+flair
```
sanchit97, line 3: Adding a note here just in case PyTorch versions break this requirement in the future.

tests/sample_inputs/attack_from_file.py
```diff
@@ -6,6 +6,4 @@ def Attack(model):
     search_method = textattack.search_methods.GreedyWordSwapWIR()
     transformation = textattack.transformations.WordSwapRandomCharacterSubstitution()
     constraints = []
-    return textattack.shared.Attack(
-        goal_function, constraints, transformation, search_method
-    )
+    return textattack.Attack(goal_function, constraints, transformation, search_method)
```
sanchit97, line 9: Doesn't flake8 handle this?

tests/test_command_line/test_attack.py
```diff
@@ -20,7 +20,6 @@
             "textattack attack --model cnn-imdb "
             "--attack-from-file tests/sample_inputs/attack_from_file.py^Attack "
             "--num-examples 2  --num-examples-offset 18 --attack-n "
-            "--shuffle=False"
```
sanchit97, line 23: Why do we disable shuffle? Do we always perform tests on a fixed sentences of the dataset? Isn't it better to test from random sentences everytime?
Make testing deterministic(by fixed) vs improving coverage (by random)

textattack/transformations/word_insertions/word_insertion_masked_lm.py
```diff
@@ -20,6 +20,9 @@ class WordInsertionMaskedLM(WordInsertion):
             you can skip this argument as the correct tokenizer can be infered from the name. However, if you're passing the actual model, you must
             provide a tokenizer.
         max_length (int): the max sequence length the masked language model is designed to work with. Default is 512.
+        window_size (int): The number of surrounding words to include when making top word prediction.
```
sanchit97, line 23: I am Out of Loop here, why is this window size necessary?

textattack/search_methods/search_method.py
```diff
@@ -62,6 +58,14 @@ def is_black_box(self):
         model's internal states."""
         raise NotImplementedError()
 
+    def get_victim_model(self):
```
sanchit97, line 61: This is a great check!

textattack/trainer.py
```diff
@@ -0,0 +1,884 @@
+import collections
```
sanchit97, line 1: Is there a test for this file (adv training in general) during builds?

textattack/models/helpers/utils.py
```diff
@@ -10,17 +10,14 @@
...
-    model_folder_path = utils.download_if_needed(model_folder_path)
     # Take the first model matching the pattern *model.bin.
     model_path_list = glob.glob(os.path.join(model_folder_path, "*model.bin"))
     if not model_path_list:
         raise FileNotFoundError(
             f"model.bin not found in model folder {model_folder_path}."
         )
     model_path = model_path_list[0]
-    state_dict = torch.load(model_path, map_location=utils.device)
+    state_dict = torch.load(model_path, map_location=torch.device("cpu"))
```
sanchit97, line 22: Just to clarify, the reason we do this, is to always load params on the CPU and then transfer to GPU as when required?

sanchit97 (approved):
Great work @jinyongyoo !

qiyanjun:
@jinyongyoo can you check the conflict files: 

tests/test_command_line/test_attack.py
textattack/commands/attack/attack_args.py
textattack/commands/attack/attack_args_helpers.py
textattack/commands/peek_dataset.py
textattack/commands/train_model/run_training.py
textattack/shared/word_embedding.py

Then we merge!

tests/test_command_line/test_train.py
```diff
@@ -27,9 +27,9 @@ def test_train_tiny():
     train_acc = re.findall(r"Train accuracy: (\S+)", stderr)
     assert train_acc
     train_acc = float(train_acc[0][:-1])  # [:-1] removes percent sign
-    assert train_acc > 50
+    assert train_acc > 60 and train_acc < 80
```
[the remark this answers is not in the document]
qiyanjun: @jinyongyoo I think this is a legitimate concern!

textattack/models/helpers/utils.py
jinyongyoo: Yes

textattack/trainer.py
jinyongyoo: We do have one command line test. But we should add API version next.

textattack/transformations/word_insertions/word_insertion_masked_lm.py
jinyongyoo: It's an option if we do not want to look at the entire text to make the prediction for one [MASK] token.

textattack/attack_results/failed_attack_result.py
```diff
@@ -1,8 +1,3 @@
-"""
```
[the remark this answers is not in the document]
jinyongyoo: Yes

tests/test_command_line/test_train.py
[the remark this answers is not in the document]
jinyongyoo: I think I included it as part of sanity checking since we only train for one epoch. We can choose to remove it.

PR #443, opened 2021-04-09
Fix file redirection bug

cogeid:
# What does this PR do?

## Summary
Automatically change color method to file when redirecting output


qiyanjun:
@sanchit97 please review

qiyanjun:
@jinyongyoo please review.. then we merge

textattack/loggers/attack_log_manager.py
```diff
@@ -19,15 +19,16 @@ def __init__(self):
 
     def enable_stdout(self):
         self.loggers.append(FileLogger(stdout=True))
+        return None
```
jinyongyoo, line 22: Why do we return `None`?

jinyongyoo, line 22: Why do we return `None`?

jinyongyoo (commented):
Seems like this PR is missing `--nocolor` flag feature.

textattack/loggers/attack_log_manager.py
qiyanjun: @cogeid please respond..

cogeid:
> Seems like this PR is missing `--nocolor` flag feature.
I felt like original issue was more of a bug than a lack of feature. In no situation would the user want to output the text with "color" when redirecting it to a file, since it results in illegible outputs. So I figured it would make sense to fix the bug rather than add a "nocolor" feature.

textattack/loggers/attack_log_manager.py
cogeid: yeah i'm not sure why that is there, will be fixed in the latest commit

qiyanjun:
@cogeid your comments on --noncolor make sense! 

@jinyongyoo do you agree?

qiyanjun:
@jinyongyoo is this ready to merge? 
If not, @cogeid please fix

cogeid:
@qiyanjun @jinyongyoo I am confident that this is ready to merge.

PR #487, opened 2021-07-23
Implement Chinese Attack to textattack

cogeid:
# What does this PR do?

## Summary

This PR allows users to attack Chinese datasets using the same procedures as attacking an english dataset.

## Additions
- Added a new notebook tutorial demonstrating how to attack Chinese models and augmenting Chinese data `Example_6_ChineseAttack.ipynb`
- Added Chinese homophone character swap transformation
- Added Chinese hownet wordswap transformation
- Added a chinese homophone dictionary in the form of .txt file

## Changes
- strings.py (automatic language detection is implemented so that words can be segmented correctly for different languages.)


jxmorris12:
Hi @cogeid - this PR is looking pretty good! is it close to being finished?

textattack/shared/utils/strings.py
```diff
@@ -76,19 +79,43 @@ def words_from_text(s, words_to_ignore=[]):
...
-        words.append(word)
+
+    DetectorFactory.seed = 0
+    try:
+        if s:
+            if detect(s) == "zh-cn" or detect(s) == "ko":
+                seg_list = jieba.cut(s, cut_all=False)
+                s = " ".join(seg_list)
+            else:
+                s = " ".join(s.split())
```
Hanyu-Liu-123, line 90: Should try/except be made in this session instead to avoid repeating the same code blocks?

docs/2notebook/2_Constraints.ipynb
```diff
@@ -79,27 +79,57 @@
   {
    "cell_type": "code",
    "execution_count": 1,
-   "metadata": {
```
Hanyu-Liu-123, line 82: I think the change in this notebook is by mistake? We shouldn't delete its original ouputs

docs/2notebook/4_Custom_Datasets_Word_Embedding.ipynb
```diff
@@ -34,6 +34,101 @@
   {
    "cell_type": "code",
    "execution_count": 1,
+   "metadata": {},
```
Hanyu-Liu-123, line 37: This one as well. I don't think we should include the changes here

docs/2notebook/Example_6_Chinese Attack.ipynb
```diff
@@ -0,0 +1,590 @@
+{
```
Hanyu-Liu-123, line 1: I think this notebook provides a really good illustration of using the Chinese attack. Providing more explanations and fixing up the formats would definitely make this better. For example, in the last section, there is no need to import `SwapRandomCharacterDeletion, WordSwapQWERTY, CompositeTransformation` since they are not used, which may confuse new users.

qiyanjun (changes requested):
have the changes Hanliu recommended been in?

cogeid:
> have the changes Hanliu recommended been in?

I will message you as soon as I finish those!

docs/2notebook/Example_6_Chinese Attack.ipynb
qiyanjun: Agreed

textattack/shared/utils/strings.py
cogeid: The reason why it had to be repeated is 

TRY:detect language:

if Chinese/korean, segment using the chinese word segmenter.
         else: use normal segmentation for germanic languages.
EXCEPTION: If language cannot be detected (sometimes language detectors throws an error if the input is empty for example):
          segment the sentence into words.

qiyanjun:
@Hanyu-Liu-123 can you make a final pass?

Hanyu-Liu-123:
> @Hanyu-Liu-123 can you make a final pass?

I think @cogeid is still working on editing the Chinese Attack notebooks (and reverting unwanted changes). There also seem to be issues that slow down PyTest.

cogeid (commented):
Hanyu's suggestions have been added in

wenh06:
How do you deal with the inconsistency of `jieba.cut` with `jieba.posseg.cut` (`pseg.cut`)? I'm maintaining a private Chinese version of `TextAttack`, which is also based on `OpenAttack`. In some `Constraint`, you are to provide the POS of the words. In this case, you have to use `pseg.cut` again, and sometimes the word list is different. For example:
image
The warning was previously an assertion. Is there a method or model accepting a list of Chinese words as input and do POS tagging on this list without doing `cut` again? I searched but couldn't find any.

My `jieba_tag` function and Chinese word cut function are as follows

```python
import re
from string import punctuation

import jieba
import jieba.posseg as pseg
from zhon.hanzi import punctuation as zh_punctuation, stops

_jieba_use_paddle = False

# fmt: off
_jieba_ner_mapping = {
    "nr": "PER", "PER": "PER",  # 人名
    "ns": "LOC", "LOC": "LOC",  # 地点
    "nt": "ORG", "ORG": "ORG",  # 机构
    "t": "TIME", "TIME": "TIME",  # 时间
    "nw": "WORK",  # 作品名
}
# fmt: on

def jieba_tag(s: str, tag_type: str = "pos") -> tuple:
    """ """
    assert tag_type.lower() in [
        "pos",
        "ner",
    ], f"""tag_type must be "pos" or "ner", but got {tag_type}"""
    purifier = lambda w: len(re.sub(f"[{punctuation+zh_punctuation}\s]+", "", w)) > 0
    words, tags = [], []
    for item in pseg.cut(s.strip().replace("\ufeff", ""), use_paddle=_jieba_use_paddle):
        if not purifier(item.word):
            continue
        words.append(item.word)
        if tag_type.lower() == "ner":
            tags.append(
                normalize_ner_tag(
                    _jieba_ner_mapping.get(item.flag, "OTHER"),
                )
            )
        else:  # pos
            tags.append(normalize_pos_tag(item.flag))
    res = {"words": words, "tags": tags}
    return res

def words_from_text_cn(s: str, words_to_ignore: list = []) -> list:
    """ """
    words = jieba.cut(s.strip().replace("\ufeff", ""), use_paddle=_jieba_use_paddle)
    purifier = (
        lambda w: len(
            re.sub(
                f"([{punctuation+zh_punctuation}\s]+)|{'|'.join(words_to_ignore)}",
                "",
                w.strip(),
            )
        )
        > 0
    )
    words = list(filter(purifier, words))
    return words
```
the 'normalize' methods (or unify methods) are not included, as they are long and boring.

By the way, I modified your `words_from_text` function as
```python
def words_from_text(s: str, words_to_ignore: list = []) -> list:
    """ """
    homos = "ɑЬϲԁе𝚏ɡհіϳ𝒌ⅼｍոорԛⲅѕ𝚝սѵԝ×уᴢ"
    exceptions = "'-_*@"
    filter_pattern = homos + "'\-_\*@"
    filter_pattern = f"[\w{filter_pattern}]+"
    words = []
    for word in s.split():
        # Allow apostrophes, hyphens, underscores, asterisks and at signs as long as they don't begin the word.
        word = word.lstrip(exceptions)
        filt = [w.lstrip(exceptions) for w in re.findall(filter_pattern, word)]
        words.extend(filt)
    words = list(filter(lambda w: w not in words_to_ignore + [""], words))
    return words
```
which I think gives better results:
image

qiyanjun:
@wenh06 great inputs! 
Mind to submit revisions to this PR?

qiyanjun:
the sphix doc error seems to come from  https://github.com/sphinx-doc/sphinx/issues/10289

qiyanjun (approved):

qiyanjun:
@wenh06 how about you submit a separate PR, after I merge this into main?

PR #563, opened 2021-11-03
Add API docstrings for back translation

cogeid:
Added API docstrings for back translation.

Also provides a temporary fix to pytest caused by a known issue from tensorflow.
the issue can be seen here
https://github.com/huggingface/transformers/issues/14265
https://github.com/huggingface/transformers/issues/14266
to solve the issue tensorflow 2.5.0 is installed instead of 2.6.0

Also added test for back translation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ extras["tensorflow"] = [
     "tensorflow_hub",
     "tensorflow_text>=2",
     "tensorboardX",
-    "tensorflow-estimator==2.6.0",
+    "tensorflow-estimator==2.5.0",
 ]
 
 extras["optional"] = [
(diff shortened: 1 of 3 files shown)


qiyanjun:
@cogeid why we need to change setup for a docstring update? 

` "tensorflow>=2",
     "tensorflow==2.5.0",
     "tensorflow_hub",
     "tensorflow_text>=2",
     "tensorboardX",
     "tensorflow-estimator==2.6.0",
     `"tensorflow-estimator==2.5.0",`

qiyanjun:
@cogeid where is the pytest fix?

qiyanjun (approved):

PR #567, opened 2021-11-07
add language options

cogeid:
# What does this PR do?

## Summary
This PR adds language options to transformations.

## Additions

- french_recipe.py
- spanish_recipe.py

two basic recipes that uses WordNet transformation to transform French/Spanish synonyms, and uses change_name and change_location transformations.

## Changes

- data.py
- word_swap_change_name.py
- word_swap_change_location.py

Spanish and French data has been added to data.py to support multilingual name and location transformation. The user can now add a parameter to these two transformations such as
`transformation = word_swap_change_name(language = "spanish")` 
which was previously not possible

- docs/2notebook/Example_4_CamemBERT.ipynb

This multilingual tutorial has been updated to use the new french recipe

**Here is a snippet of code that demonstrates the changes made**

Screen Shot 2021-11-12 at 6 02 45 AM


qiyanjun:
@cogeid please add a thorough PR description 

@cogeid please replace test_augment_api.py with the master branch’s version. 

@cogeid please add docstrign API use case in each recipe..

cogeid:
> @cogeid please add a thorough PR description
> 
> @cogeid please replace test_augment_api.py with the master branch’s version.

I have completed the first two requests. Was late last night, sorry about that!

> @cogeid please add docstrign API use case in each recipe..
I am hesitant on doing this because currently none of transformations have docstrings demonstrating API usage, I feel it's better if we add the whole module with consistent docstrings in one update rather than one by one. I will work on the docstrings this week for transformation.

qiyanjun:
@alexander-zap mind to help me review this PR?

qiyanjun:
@Hanyu-Liu-123 mind helping me review this?

Hanyu-Liu-123:
> @Hanyu-Liu-123 mind helping me review this?

Will do today!

Hanyu-Liu-123 (commented):
I haven't tested this out but I think we may need to change one more thing:

`WordSwapChangeLocation` and `WordSwapChangeName` use flair ner tagger to search for indices that can be changed to alternative locations/names. The tagger is implemented [here](https://github.com/QData/TextAttack/blob/49c94fb59bc60cae9e5e39d1a8a78f932a9fe9ed/textattack/shared/attacked_text.py#L164), and uses the default flair [ner](https://github.com/QData/TextAttack/blob/49c94fb59bc60cae9e5e39d1a8a78f932a9fe9ed/textattack/shared/attacked_text.py#L173) model. If we're to transform texts in Spanish or French, we may want to use ner model for the respective language (`ner-french` and `ner-spanish` as you can find on their [website](https://huggingface.co/flair?sort=downloads)) for better performance. 

Otherwise it looks good to me.

cogeid:
> I haven't tested this out but I think we may need to change one more thing:
> 
> `WordSwapChangeLocation` and `WordSwapChangeName` use flair ner tagger to search for indices that can be changed to alternative locations/names. The tagger is implemented [here](https://github.com/QData/TextAttack/blob/49c94fb59bc60cae9e5e39d1a8a78f932a9fe9ed/textattack/shared/attacked_text.py#L164), and uses the default flair [ner](https://github.com/QData/TextAttack/blob/49c94fb59bc60cae9e5e39d1a8a78f932a9fe9ed/textattack/shared/attacked_text.py#L173) model. If we're to transform texts in Spanish or French, we may want to use ner model for the respective language (`ner-french` and `ner-spanish` as you can find on their [website](https://huggingface.co/flair?sort=downloads)) for better performance.
> 
> Otherwise it looks good to me.

I will look into the flair issue.

alexander-zap:
@qiyanjun I like the usage of the new transformation classes which can be configured to be of any language. That's great!

But I would recommend for the future to not hard-code language-specific recipe classes (e.g., `FrenchRecipe`, `SpanishRecipe`, ...) because this requires creating a new class for each language. 

For the future instead I would suggest to have something similar to a `LanguageResourceProvider` for each language which manages (here it can be hard-coded) all language-specific resources and can on runtime create the required transformations. These can then be used in a language agnostic `Recipe` class which is then able to query the required transformations based on the wished language.

qiyanjun:
@cogeid please help add some tests on the functions you added. 

@alexander-zap thanks for the reviews

textattack/transformations/word_swaps/word_swap_change_location.py
```diff
@@ -43,13 +43,19 @@ def __init__(self, n=3, confidence_score=0.7, **kwargs):
...
         self.confidence_score = confidence_score
+        self.language = language
 
     def _get_transformations(self, current_text, indices_to_modify):
         words = current_text.words
         location_idx = []
-
+        if self.language == "en":
+            model_name = "ner"
+        elif self.language == "fra" or self.language == "french":
```
qiyanjun, line 53: so here we do not need something like follows? :

`  if self.language == "esp" or self.language == "spanish":`

textattack/transformations/word_swaps/word_swap_change_name.py
```diff
@@ -40,13 +41,20 @@ def __init__(
...
         self.confidence_score = confidence_score
+        self.language = language
 
     def _get_transformations(self, current_text, indices_to_modify):
         transformed_texts = []
+        if self.language == "en":
+            model_name = "ner"
+        elif self.language == "fra" or self.language == "french":
+            model_name = "flair/ner-french"
+        else:
```
qiyanjun, line 52: so here we do not need something like follows? :

`  if self.language == "esp" or self.language == "spanish":`

cogeid: We do not need to do self.language == "esp" or self.language == "spanish" because the model we use for spanish is a multilingual model from flair(https://huggingface.co/flair/ner-multi-fast) includes german dutch spanish and english. It however only "kind of works" for related languages such as french, so french is the only language that needs a separated model. We so far do not have support for german and dutch yet, but if we were to add them they should work without changing the logic here.

qiyanjun: I see. this works!

qiyanjun (approved):

PR #582, opened 2021-11-17
Fix extra quotation marks issue

Hanyu-Liu-123:
# What does this PR do?

## Summary
When augmenting from CSV files, we use the CSV sniffer to try automatically inferring the correct CSV format. This works most of the time but does not function properly when there're more than 2 double quotation marks within a line, due to the fact that double quotation marks are recognized as escape characters by CSV sniffer. 

The workaround first goes through the input CSV file, replaces double quotation marks with single quotation marks, and then creates a temporary CSV file to store the modified inputs. We need to create the temp file because we couldn't directly provide the modified inputs (list of strings) to the CSV sniffer, so the modified lines need to be saved to a CSV file and that file will be read by the sniffer.

Users will be warned on which lines of the original input file are modified.

## Additions
- csv_reader_workaround()

### DIFF
--- a/textattack/commands/augment_command.py
+++ b/textattack/commands/augment_command.py
@@ -137,9 +137,12 @@ class AugmentCommand(TextAttackCommand):
                     raise OSError(
                         f"Outfile {args.output_csv} exists and --overwrite not set."
                     )
+
+            csv_reader_workaround(args.input_csv)
+
             # Read in CSV file as a list of dictionaries. Use the CSV sniffer to
             # try and automatically infer the correct CSV format.
-            csv_file = open(args.input_csv, "r")
+            csv_file = open("temp.csv", "r")
             dialect = csv.Sniffer().sniff(csv_file.readline(), delimiters=";,")
             csv_file.seek(0)
             rows = [
@@ -148,6 +151,7 @@ class AugmentCommand(TextAttackCommand):
                     csv_file, dialect=dialect, skipinitialspace=True
                 )
             ]
+            os.remove("temp.csv")
             # Validate input column.
             row_keys = set(rows[0].keys())
             if args.input_column not in row_keys:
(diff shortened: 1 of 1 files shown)


Hanyu-Liu-123:
The issue I was trying to solve with the PR is that unwanted quotation marks will be added to the output csv file if there’re double quotation marks within a sentence. It comes in 2 folds:

suppose we have the below csv file：

> text, label
> Today is a "good" day, 1

it will return something like:

> text, label
> Today is a ""fantastic"" day, 1

because double quotation mark is the escape character for csv files, a double quotation mark needs a double quotation mark when writen to the csv file.


2. the second problem is when the sentence in the  input file is wrapped in double quotation marks like the following:

> text, label
> "Today is a "good" day", 1

it will return something like:

> text, label
> "Today is a fantastic"" day""", 1

which is due to how the csv sniffer handles the input file. It has this minor glitch when double quotation marks are found inside a sentence already wrapped in double quotation marks. When it reads `"Today is a "good" day"` from the input file, it recognizes it as  `Today is a good" day"` (the first 2 quotation marks are removed and the rest remains). Then after the augmentation extra quotation marks are added to those remaining quotation marks so the output is `"Today is a fantastic"" day"""`

I wasn’t able to find a way to fix the csv sniffer, so I thought  we could just preprocess the input file by replacing the double quotation marks with single quotation marks. But this turned out to be a messy solution because we need to create a temp file to store the processed input for the csv sniffer, overwriting the original input file may be risky.

donggrant:
Some more cases with the bug:
<html>
<body>
<b style="font-weight:normal;" id="docs-internal-guid-f2744ef6-7fff-f4d4-b070-51a2685af317"><div dir="ltr" style="margin-left:0pt;" align="left">

Input | Output
-- | --
Today is a good day | Today is a good day
Today is a "good" day | "Today is a ""good"" day"
“Today is a good day” | Today is a good day
""Today is a "good" day"" | "Today is a ""good"" day"""""
"Today is a "good" day" | "Today is a good"" day"""
Today is a "good" day" | "Today is a ""good"" day"""
Today "is a "good" day" | "Today ""is a ""good"" day"""

</div></b>
</body>
</html>

qiyanjun:
see #623 for a newer solution

PR #617, opened 2022-02-23
Fix Pytest Issues

Hanyu-Liu-123:
# What does this PR do?

## Summary
The recent failed test cases seem to come from 3 problems:

 - [x] On this [line](https://github.com/QData/TextAttack/blob/2ee54962435411a55c042bd100d44826377c866e/textattack/transformations/word_swaps/word_swap_wordnet.py#L28), `wordnet.langs()` is accessed and results in a lookup error since the package`omw-1.4` is missing. By downloading `omw-1.4` the issue was solved.
- [x] Assertion Error in one of the tests:  `tests/test_command_line/test_attack.py::test_command_line_attack[attack_from_file-textattack attack --model cnn-imdb --attack-from-file tests/sample_inputs/attack_from_file.py^Attack --num-examples 2  --num-examples-offset 18 --attack-n -tests/sample_outputs/run_attack_from_file.txt]`. Specifically, different sentences are attacked when running locally or running with GitHub's testing.
- [] Notebook formatting problem causing Sphinx to raise an error

### DIFF
--- a/.github/workflows/make-docs.yml
+++ b/.github/workflows/make-docs.yml
@@ -35,4 +35,4 @@ jobs:
         pip install jupyter 'ipykernel<5.0.0' 'ipython<7.0.0' # ipykernel workaround: github.com/jupyter/notebook/issues/4050
     - name: Build docs with Sphinx and check for errors
       run: |
-        sphinx-build -b html docs docs/_build/html -W
+        sphinx-build -b html docs docs/_build/html
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ docs: FORCE ## Build docs using Sphinx.
 	sphinx-build -b html docs docs/_build/html
 
 docs-check: FORCE ## Builds docs using Sphinx. If there is an error, exit with an error code (instead of warning & continuing).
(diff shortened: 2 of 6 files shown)


Hanyu-Liu-123:
More detail on the second issue (`tests/test_command_line/test_attack.py::test_command_line_attack[attack_from_file-textattack attack --model cnn-imdb --attack-from-file tests/sample_inputs/attack_from_file.py^Attack --num-examples 2 --num-examples-offset 18 --attack-n -tests/sample_outputs/run_attack_from_file.txt]`):

In the requirement docs we do not specify which version of `dataset` library to use. We use the `dataset` library from Hugginface to load datasets to be attacked (including the imdb dataset used in the above test). With their recent update, the order of the sentences were changed when using `dataset`, so different sentences were attacked when running though Github's testing (which uses the latest version of `dataset`) and running locally (which uses previous versions). This results in different results and the assertion error. 

To solve this, we could either specific `dataset` version, or update the desired test output so that the two results match.

qiyanjun:
@Hanyu-Liu-123 let us use old version of dataset library for now

qiyanjun (approved):

PR #619, opened 2022-02-28
Update datasets version

Hanyu-Liu-123:
# What does this PR do?

## Summary
datasets version needs to be lower than 1.16 to pass existing tests

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ filelock
 language_tool_python
 lemminflect
 lru-dict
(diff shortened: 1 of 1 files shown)


Hanyu-Liu-123:
@qiyanjun The version of `datasets` in #617 is still too high. Downgrading to version 1.15 solves the problem.

jxmorris12:
thank you @Hanyu-Liu-123  :)

PR #623, opened 2022-03-04
Fixed quotation issues

donggrant:
# What does this PR do?
Fixes the bug in the Augmenter Command class where extra quotation marks are added to the output csv files. This PR avoids using any changes to the input file or having to create a temporary file

## Summary
Instead of changing how the csv reader works, an easier solution is to mark where quotes and commas occur within the text field while the data is being read into the DictReader(). Then before augmenting, commas and quotes are replaced back into the text value. This avoids the repeated quotation issue and the glitch where the first two quotes are removed by the csv reader. After the augmentation process, the output file is altered to remove any additional markings from the preprocessing.

### DIFF
--- a/textattack/commands/augment_command.py
+++ b/textattack/commands/augment_command.py
@@ -140,14 +140,36 @@ class AugmentCommand(TextAttackCommand):
             # Read in CSV file as a list of dictionaries. Use the CSV sniffer to
             # try and automatically infer the correct CSV format.
             csv_file = open(args.input_csv, "r")
+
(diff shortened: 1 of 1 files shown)


qiyanjun:
@Hanyu-Liu-123 please review

Hanyu-Liu-123 (approved):
Looks good to me! I think we could also add this solution to the csv logger for attack results since it most likely has the same issue with quotation marks. 

I think it would also be helpful if we provide a warning for users if there is `/` in the original sentence, since it will be deleted in the ouput.

PR #639, opened 2022-05-04
Extend qa model

srujanjoshi:
# What does this PR do?

This PR extends TextAttack to work for QA tasks. Specifically, the SQuAD dataset. It also adds a new squad attack recipe and a new custom goal function called single_word_difference.

## Summary 
A new modelwrapper was created to handle QA model inputs and outputs. It is called huggingface_qa_modelwrapper.py. The new attack recipe is named qa_squad_attack_guan_2020.py and is loosely based on this research paper: [https://web.stanford.edu/class/archive/cs/cs224n/cs224n.1204/reports/default/report23.pdf](https://github.com/QData/TextAttack/pull/url)
Lastly, the single_word_difference goal function was added to determine when an answer span is different enough. The implementation is similar to the NonOverlapping goalf function, except the goal is satisfied when there is a word difference greater than 0.

### DIFF
--- a/textattack/attack.py
+++ b/textattack/attack.py
@@ -408,7 +408,11 @@ class Attack:
             example, (str, OrderedDict, AttackedText)
         ), "`example` must either be `str`, `collections.OrderedDict`, `textattack.shared.AttackedText`."
         if isinstance(example, (str, OrderedDict)):
-            example = AttackedText(example)
+            if "context" in example:
+                altered_context = AttackedText(example["context"])
+                example["context"] = altered_context
+            else:   
+                example = AttackedText(example)
 
         assert isinstance(
             ground_truth_output, (int, str)
--- a/textattack/attack.py
+++ b/textattack/attack.py
@@ -408,7 +408,7 @@ class Attack:
             example, (str, OrderedDict, AttackedText)
         ), "`example` must either be `str`, `collections.OrderedDict`, `textattack.shared.AttackedText`."
 
-        if isinstance(example, (str, OrderedDict)):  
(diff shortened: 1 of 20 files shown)


jxmorris12 (changes requested):
@srujanjoshi - thanks for contributing!

- There are a lot of erroneous file changes, like newlines, and some commented-out print statements. Can you remove these?
- I think the HuggingFaceQAModelWrapper should depend on the `HuggingFaceModelWrapper` that we already have. At first glance, there's a good amount of duplicated code between the two. We try to prefer polymorphism (in this case, inheritance) to duplicating code. That way, if we have to fix something (for example, update to accommodate for changes in the HuggingFace APIs) then we only have to fix it in one place, not two.

I might be wrong about the inheritance so if I am please just let me know. I read the code somewhat quickly.

In any case, please give a thorough code review of yourself before requesting one from me again. Ask yourself "is this the simplest and most elegant way to do this? Does this really need to be here?" for every line. 

This is a great addition, btw, and something I've been wanting for a really long time! I can't wait to test it out!

Hanyu-Liu-123:
The current failing tests stem from this [line](https://github.com/srujanjoshi/TextAttack/blob/b86d0f840b6f13b1eb8d0f17b3fceeeca62b08ed/textattack/shared/attacked_text.py#L171), which was fixed in pull request #644. @srujanjoshi Could you approve my pull request at your personal branch that merges the master branch into this qa model branch? Thanks!

jxmorris12:
@srujanjoshi let us know if you want to continue pursuing this.

Issue #697, opened 2022-10-25
Attacking T5 model doesn't work


poojithat512:
<img width="1075" alt="Screen Shot 2022-10-25 at 8 20 08 AM" src="https://user-images.githubusercontent.com/90569109/197785272-cde74e2f-a3fc-4753-b71a-a4443c8f220c.png">

jxmorris12:
Hopefully I fixed this today with #700 :) can you try and pull the current code and let me know if your issue is fixed?

jxmorris12:
Pretty sure this was fixed a few days ago! Please reopen if you're still having issues

PR #701, opened 2022-11-01
Fix bugs with t5

jxmorris12:
Now this command should work: `textattack attack --model t5-en-de --recipe seq2sick --num-examples 5`. This is still in reference to #697.


jxmorris12:
nice job jack

PR #709, opened 2022-12-13
Fix the problem of text output from T5 model

plasmashen:
To deal with the text output from the T5 model, where the array of text output cannot be converted into a tensor. Therefore, directly processing the numpy array is better.


jxmorris12:
Hi @plasmashen -- looks like your code works for your case but not in the general case. The tests caught an error where the changes (maybe the call to `.ravel()`?) break the output shape:

```bash
    goal_function_result, _ = self.goal_function.init_attack_example(
  File "/home/runner/work/TextAttack/TextAttack/textattack/goal_functions/goal_function.py", line 68, in init_attack_example
    result, _ = self.get_result(attacked_text, check_skip=True)
  File "/home/runner/work/TextAttack/TextAttack/textattack/goal_functions/goal_function.py", line 79, in get_result
    results, search_over = self.get_results([attacked_text], **kwargs)
  File "/home/runner/work/TextAttack/TextAttack/textattack/goal_functions/goal_function.py", line 96, in get_results
    model_outputs = self._call_model(attacked_text_list)
  File "/home/runner/work/TextAttack/TextAttack/textattack/goal_functions/goal_function.py", line 218, in _call_model
    outputs = self._call_model_uncached(uncached_list)
  File "/home/runner/work/TextAttack/TextAttack/textattack/goal_functions/goal_function.py", line 189, in _call_model_uncached
    assert len(inputs) == len(
AssertionError: Got 2 outputs for 1 inputs
```

Can you think of a way to fix this?

jxmorris12 (approved):

PR #734, opened 2023-05-15
Fixed syntax and import issues in the example of Attack API

eldorabdukhamidov:
# What does this PR do?

## Summary
*This PR fixes several issues in the example of an Attack API. Without the fixes, a user is not able to run the example code*

## Additions
- *I added several package imports such as _WordSwapEmbedding_, GreedyWordSwapWIR as the example shows **package not found** issue. *

## Changes
- *I updated the call of Attack API in line 73*


jxmorris12 (approved):
thanks for fixing the docs

PR #752, opened 2023-10-05
Consistent word swap

k-ivey:
# What does this PR do?

## Summary
This PR adds a new `consistent` parameter to `WordSwapChangeLocation` and `WordSwapChangeName` so that all original instances of the same word get updated to the same new word. This allows for consistency when the semantics of the words matter. For example,  "Alice likes Bob. Alice does not like Eve." could be transformed into "Kate likes Bob. Kate does not like Eve" if `consistent = True`. Otherwise, the text could be transformed into "Alice likes Bob. Kate does not like Eve." which alters the semantics of the sentence. 

## Additions
- Added `consistent` parameter to `WordSwapChangeLocation` and `WordSwapChangeName`. 

## Changes
- `WordSwapChangeLocation` has been updated to only make one modification to the original text for marginal  performance gains. 
- `WordSwapChangeLocation` has been updated to correctly capitalize locations consisting of more than one word to match the naming convention used in `textattack/shared/data.py`. This change allows for replacement words for such multi-word locations to be found. 

## Checklist
- [ x ] The title of your pull request should be a summary of its contribution.
- [ x ] Please write detailed description of what parts have been newly added and what parts have been modified. Please also explain why certain changes were made.
- [ x ] All public methods must have informative docstrings that work nicely with sphinx. For new modules/files, please add/modify the appropriate `.rst` file in `TextAttack/docs/apidoc`.'

### DIFF
--- a/tests/test_transformations.py
+++ b/tests/test_transformations.py
@@ -33,6 +33,36 @@ def test_word_swap_change_location():
     assert entity_original == entity_augmented
 
 
+def test_word_swap_change_location_consistent():
+    from flair.data import Sentence
+    from flair.models import SequenceTagger
+
+    from textattack.augmentation import Augmenter
+    from textattack.transformations.word_swaps import WordSwapChangeLocation
+
+    augmenter = Augmenter(transformation=WordSwapChangeLocation(consistent=True))
+    s = "I am in New York. I love living in New York."
+    s_augmented = augmenter.augment(s)
+    augmented_text = Sentence(s_augmented[0])
+    tagger = SequenceTagger.load("flair/ner-english")
+    original_text = Sentence(s)
+    tagger.predict(original_text)
+    tagger.predict(augmented_text)
+
+    entity_original = []
+    entity_augmented = []
+
+    for entity in original_text.get_spans("ner"):
+        entity_original.append(entity.tag)
+    for entity in augmented_text.get_spans("ner"):
+        entity_augmented.append(entity.tag)
+
+    print(entity_original)
+
+    assert entity_original == entity_augmented
+    assert s_augmented[0].count("New York") == 0
+
+
 def test_word_swap_change_name():
     from flair.data import Sentence
     from flair.models import SequenceTagger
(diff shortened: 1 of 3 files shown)


qiyanjun:
@k-ivey  please add test function showing the revised parts are working as intended.

k-ivey:
@qiyanjun I added two new tests in `tests/test_transformations.py`:

- `test_word_swap_change_location_consistent()` 
- `test_word_swap_change_name_consistent()`

Both tests modify an initial text with duplicated location or name and verify that the transformed text contains no instances of the initial location or name. 

Since the PyTest action appears to be failing due to memory issues, I'll introduces changes to the workflow to increase the swap size of the machine.

qiyanjun:
@k-ivey please sync your branch with the main

k-ivey:
@qiyanjun I synced my branch and the tests now pass

qiyanjun (changes requested):
The name change seems ok. 

But the location change needs some debugging. 

for instance: 

```

s = "I am in New York. I love living in San Diego. San Diego is better than New York"
>>> s_augmented = augmenter.augment(s)
>>> print(s_augmented)
['I am in New York. I love living in San Diego. Everett is better than New York']
>>> 
```

k-ivey:
I've looked a bit into this and it appears to be a consequence of the NER tagger along with multi-word locations. For the sentence
```
s = "I am in New York. I love living in San Diego. San Diego is better than New York"
```
The first instance of "New York" is tagged as a location. However, for the second instance of "New York", only "York" is tagged as a location. Similarly, the second instance of "San Diego" is tagged as a location, but for the first instance, only "San" is tagged as a location. I'll work some more on this, as it will be tricky to handle edge cases such as text that has "San Diego" and "San Francisco".

k-ivey:
I've made changes to address the issue. In short, the issue was that the NER tagger may not tag all words in a multi-word location (e.g. New York). To fix this, we can find all windows that the NER tagger identifies as a location, sort these by the number of words in the window, and then extend the shorter windows to see if the extended window is a location we've already seen. If so, the update the map of location to its appearances. As a concrete example:
```
s = "I am in New York. I love living in San Diego. San Diego is better than New York"
```
The locations identified by NER are [ [[3, 4], "New York"], [[9], "San"], [[11, 12], "San Diego"], [[17], "York"], where the indices are word indices in the sentence. The longest window is 2 and the shortest is 1, so we need to expand the windows of length 1 to length 2:

- For "San" at word index 9, we expand it to "in San" and "San Diego". The latter is a location that already appears, so the value for "San Diego" in the map is updated to include this expanded window. 
- For "York" at word index 17, we expand it to "New York", so the value for "San Diego" in the map is updated to include this expanded window. 

This does not handle all issues with the NER tagger. For example, in the example sentence above, if the NER tagger only tagged the first instance of "New York" as a location and did not tag either "New" or "York" as a location in the second instance, then the second instance would never be updated.

qiyanjun (approved):

qiyanjun:
It still has not totally solved the change location issues.. but better than the current version.. so it merged

```
>>> s = "I am in New York. San Diego is better than New York"
>>> s_augmented = augmenter.augment(s)
>>> print(s_augmented)
['I am in New York. San Diego is better than New Grand Island']

```

Issue #586, opened 2021-11-30
bert-attack doesn't work very well


marsssser:
Hello, author, when I attacked LSTM-IMDB with bert-attack, I would not continue the attack for a long time after the 11th data was attacked, that is, the 12th result of the attack could not be completed, I wonder if it needs a longer time to wait (more than 2 hour?). Or is there a problem with the code?

qiyanjun:
@marsssser are you using GPU?

qiyanjun:
what is the attack recipe name you used?

DavidHerel:
I have the same problem. Attack recipe that I used is the following:
`!textattack attack --recipe bert-attack --num-examples 1000 --model bert-base-uncased-mr --dataset-from-huggingface rotten_tomatoes --dataset-split test`

dangne:
I raised this issue several months ago in the slack channel, and the reason why BERT-Attack takes so long to run is due to this line https://github.com/QData/TextAttack/blob/776dfece2aab2c6e0b9015d04696528e1706246d/textattack/transformations/word_swaps/word_swap_masked_lm.py#L228
because:
1. The default value of `K` for BERT-Attack is 48 ([see in the code](https://github.com/QData/TextAttack/blob/776dfece2aab2c6e0b9015d04696528e1706246d/textattack/attack_recipes/bert_attack_li_2020.py#L36)), and the number of sub-word combinations grows exponentially w.r.t. `K`.
2. During the attack, if a word gets tokenized to 4 sub-words, the number of combinations is 48^4 = 5308416, which is huge!

As far as I remember, the implementation in this library is correct. I think there may be a fundamental flaw in the algorithm itself. However, I suggest three ways to mitigate the issue

1. Set `K` to something small (`k <= 8` is fine).
2. If you need to keep `K=48`, you can modify the code to ignore tokens that have more than three or four sub-words.
3. I also think the authors of this repo should raise a warning for users that want to use BERT-Attack as well.

qiyanjun:
@dangne thank you for the excellent comment! We plan to have a faster version of Bert-Attack based on your comment. If you are interested in doing it, please help us PR !~

marsssser:
> 

attack
--recipe
bert-attack
--model
bert-base-uncased-yelp
--num-examples
30

marsssser:
> what is the attack recipe name you used?

!textattack attack
--recipe
bert-attack
--model
bert-base-uncased-yelp
--num-examples
30

[...]

PR #765, opened 2023-12-13
Add files via upload

HuichiZhou:
# What does this PR do?
In some instances, using a candidate size of 48 can lead to significant time consumption. It is recommended to reduce this to fewer than 8. 
#586


jxmorris12:
Sorry, I don't think we want to change this reference implementation, since we are trying to mimic what they did in the BERT-Attack paper. Feel free to make the change on your own branch though.

PR #778, opened 2024-03-05
Do not use pipeline to achieve faster generation of Chinese mask repl…

liuyuyan2717:
The chinese_word_swap_masked.py calls pipline to generate masked language model substitution words, which is slow. Modified to generate Chinese mask substitution words for calling the local model


jxmorris12:
this looks good. thanks for contributing.

PR #794, opened 2024-07-07
Attack Recipe Implementation - Leap

jstzwj:
## Summary
Implements the attack method in "LEAP: Efficient and Automated Test Method for NLP Software", Xiao et al., 2023.
[](https://arxiv.org/abs/2308.11284)

## Adds:
* New attack recipe ("leap", leap_2023)
* New search method, a variant of ParticleSwarmOptimization (ParticleSwarmOptimizationLEAP). LEAP generates the initial population using Levy flight and Brownian motion. Then LEAP uses a new adaptive inertia weight update strategy for LEAP to optimize the search path in an exponentially growing text space.


qiyanjun (commented):
Thanks for porting LEAP — it's a good fit alongside the existing `PSOZang2020` recipe. I compared this port against the authors' reference implementation (`lumos0514/LEAP`, `search_methods/leap.py`) and found the mutation step in `particle_swarm_optimization_leap.py` has drifted from it in two ways. Both make the recipe silently behave more like vanilla PSO than LEAP:

1. **`_greedy_perturb` is defined but never called.** `perform_search` calls the inherited `self._perturb` (probabilistic sampling among best neighbors) instead of `self._greedy_perturb` (deterministic pick of the best neighbor), which is LEAP's actual mutation strategy per the reference code.
2. **`change_ratio` is computed against the wrong reference point.** It currently diffs the *current* candidate against the *original* input; the reference diffs the current candidate against its own `local_elites[k]`.

Suggested fix (verified it imports and the method now calls the right function):

```diff
             # Mutation based on the current change rate
             for k in range(len(population)):
-                change_ratio = initial_result.attacked_text.words_diff_ratio(
-                    population[k].attacked_text
+                change_ratio = population[k].attacked_text.words_diff_ratio(
+                    local_elites[k].attacked_text
                 )
                 # Referred from the original source code
                 p_change = 1 - 2 * change_ratio
                 if np.random.uniform() < p_change:
-                    self._perturb(population[k], initial_result)
+                    self._greedy_perturb(population[k], initial_result)
 
                 if self._search_over:
                     break
```

A few other things worth a look while this file is open (not blocking, flagging for awareness):

- `omega = []` (line 178) is declared once before the `max_iters` loop and only appended to, never reset — after iteration 0, `omega[k]` keeps reading stale values from the first iteration instead of the freshly computed ones, so the "adaptive" inertia weight stops adapting after the first pass. This one's actually present in the authors' own reference code too, so it's a faithful port, but probably worth a fix upstream as well.
- `fit_ave`/`fit_min` (lines 170-175) are computed once from the initial population and reused for all iterations; if the initial population ends up with identical scores (e.g. very short input / few WordNet synonyms), `fit_ave == fit_min` and the omega interpolation divides by zero.
- `turn_prob` here uses `softmax` (probabilities sum to 1 across the sentence) where the parent class's `_turn` expects an independent per-word Bernoulli probability (it uses `sigmoid` for that). For longer sentences this pushes every word's turn probability very low, effectively disabling the move-toward-elite step.
- The recipe's constraint list (`leap_2023.py`) omits `RepeatModification()`, which every other word-swap recipe in the repo includes. This matches the authors' own recipe file, so likely intentional/faithful, just flagging the inconsistency with house style.
- `docs/api/search_methods.rst`: the `ParticleSwarmOptimizationLEAP` section underline is 26 dashes for a 29-character title — Sphinx will warn ("title underline too short").

Happy to send a follow-up patch for the omega/fit_ave items too if useful — wanted to keep this comment scoped to the mutation-logic bugs since those most directly undercut LEAP's distinguishing behavior vs. plain PSO.

qiyanjun:
Thanks for this contribution! I've picked this up in #829, which includes the LEAP recipe/search method from this PR plus a fix for the two mutation-logic bugs noted in the review comment above (the greedy mutation step wasn't being called, and `change_ratio` was computed against the wrong reference text). Closing this one in favor of #829 — happy to continue any further discussion there.

Issue #796, opened 2024-07-16
Run demo error


nextdoorUncleLiu:
**Describe the bug**
https://textattack.readthedocs.io/en/latest/0_get_started/command_line_usage.html
I followed the demo in this link, but unfortunately an error message appeared in the first step.

**To Reproduce**
1. add examples.csv
2. Run command line:
```
textattack-test % textattack augment --input-csv examples.csv --output-csv output.csv  --input-column text --recipe eda --pct-words-to-swap .1 \
--transformations-per-example 2 --exclude-original
```

**Expected behavior**
It runs normally as the official website case

**Screenshots or Traceback**
![image](https://github.com/user-attachments/assets/bc2aa4a4-7637-4fa0-8529-0b5d140de669)

**System Information (please complete the following information):**
 - OS: Mac
 - python: 3.8
 - Textattack Version: 0.3.10

bterrific2008:
Looks like the issue has to do with the EDA Augmentation recipe. The EDA constructor doesn't have a catch for keyword arguments which causes the TypeError when `high_yield` is passed to that recipe.

I think the fix would be passing the keywords to the EDA recipe's composite augmenters. Something like this:

```python
def __init__(self, pct_words_to_swap=0.1, transformations_per_example=4, **kwargs):
        assert 0.0 <= pct_words_to_swap <= 1.0, "pct_words_to_swap must be in [0., 1.]"
        assert (
            transformations_per_example > 0
        ), "transformations_per_example must be a positive integer"
        self.pct_words_to_swap = pct_words_to_swap
        self.transformations_per_example = transformations_per_example
        n_aug_each = max(transformations_per_example // 4, 1)

        self.synonym_replacement = WordNetAugmenter(
            pct_words_to_swap=pct_words_to_swap,
            transformations_per_example=n_aug_each,
            **kwargs,
        )
        self.random_deletion = DeletionAugmenter(
            pct_words_to_swap=pct_words_to_swap,
            transformations_per_example=n_aug_each,
            **kwargs,
        )
        self.random_swap = SwapAugmenter(
            pct_words_to_swap=pct_words_to_swap,
            transformations_per_example=n_aug_each,
            **kwargs,
        )
        self.random_insertion = SynonymInsertionAugmenter(
            pct_words_to_swap=pct_words_to_swap,
            transformations_per_example=n_aug_each,
            **kwargs,
        )
```

This should resolve the issue. I'll test it out later this week and open a PR if all looks good.

[...]

PR #797, opened 2024-07-24
Fix bug where EDA augmentation recipe would not accept default Augmentation arguments

bterrific2008:
# What does this PR do?

## Summary

This PR resolves a bug where using the EDA augmentation recipe would not work with the textattack augmentation cli. This issue was because the EDA recipe did not accept the default arguments defined in the Augmentation superclass. This change resolves this issue by accepting a var-keyword parameter, which is then passed to the component Augmenter objects used by the EDA recipe.

This PR resolved the bug raised in #796, which found that running the EDA recipe from the textattack augmentation cli would raise a ``TypeError: __init__() got an unexpected keyword argument `high_yield` ``

## Changes

- Changes EDA recipe to accept a var-keyword parameter, which is passed to the component Augmenter objects

### DIFF
--- a/textattack/augmentation/recipes.py
+++ b/textattack/augmentation/recipes.py
@@ -37,7 +37,7 @@ class EasyDataAugmenter(Augmenter):
     https://arxiv.org/abs/1901.11196
     """
 
-    def __init__(self, pct_words_to_swap=0.1, transformations_per_example=4):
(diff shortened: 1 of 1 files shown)


jxmorris12 (approved):
Thanks! This is really helpful.

PR #829, opened 2026-08-12
Add LEAP attack recipe (fixes mutation bugs from #794)

qiyanjun:
## Summary

Adopts #794 (LEAP: Efficient and Automated Test Method for NLP Software, [arXiv:2308.11284](https://arxiv.org/abs/2308.11284)) — adds the `LEAP2023` attack recipe and `ParticleSwarmOptimizationLEAP` search method — with a series of correctness fixes and cleanups found via review against the authors' reference implementation ([lumos0514/LEAP](https://github.com/lumos0514/LEAP)):

**Correctness fixes:**
- Mutation step now calls LEAP's greedy mutation instead of silently falling back to the inherited, probabilistic `_perturb`; `change_ratio` is computed against each particle's own local elite instead of the original input, matching the reference implementation. (Without this, `leap`'s mutation step behaved identically to vanilla PSO, losing LEAP's main algorithmic distinction.)
- Fixed elite aliasing: `global_elite`/`local_elites` could end up being the *same objects* as entries in `population` (only the list, not its `PopulationMember` elements, was ever copied), so an in-place mutation could silently corrupt a tracked elite. This affects the shared base class too (also used by `PSOZang2020`).
- Guarded LEAP's adaptive inertia-weight interpolation against a zero/negative denominator, reset its per-iteration `omega` list (was accumulating stale values from iteration 0 onward, silently disabling the "adaptive" part of the algorithm after the first iteration), and capped retries in its Levy-flight rejection sampling (was an unbounded `while True`).
- Fixed a Sphinx "title underline too short" doc warning.

**Refactor:** factored `ParticleSwarmOptimization`'s per-iteration deltas into overridable hook methods (`_initialize_velocities`, `_pre_iteration_setup`, `_compute_omega`, `_compute_turn_prob`, `_compute_change_ratio`), matching the hook pattern `GeneticAlgorithm`/`AlzantotGeneticAlgorithm` already use in this codebase. `ParticleSwarmOptimizationLEAP` now overrides only these hooks plus `_perturb`, instead of duplicating a ~155-line copy of `perform_search` that had already silently drifted from the base class. Also replaced a hand-rolled `softmax` with `scipy.special.softmax` and cached the alpha-invariant constants in the Levy-flight sampler.

**Tests & docs:**
- `tests/test_attack_recipes.py`: structural test confirming `LEAP2023`'s search method shares lineage with `PSOZang2020`'s (same base class, hyperparameter defaults, goal function) while differing in transformation/constraints as intended, plus a functional test running both end-to-end as a regression guard.
- `tests/benchmark_leap_vs_pso.py`: a manual (not CI-run) benchmark comparing `LEAP2023` against a WordNet-transformation variant of vanilla PSO — holding transformation/constraints/hyperparameters fixed so the comparison isolates the search algorithm rather than conflating it with `PSOZang2020`'s different candidate-word pool (HowNet vs. WordNet). See **Benchmark results** below.
- Cross-linked `LEAP2023` ↔ `PSOZang2020` in both recipes' docstrings, the search method's docstrings, `docs/3recipes/attack_recipes.rst`, and the README recipe table (LEAP was missing from it entirely). Added `CHANGELOG.md` (none existed).

Closes #794.

## Benchmark results

Run against `cnn-ag-news` on the AG News test set (seed 765); full methodology and caveats in `LEAP2023`'s docstring and `tests/benchmark_leap_vs_pso.py`.

| | Unrestricted budget (n=20) | 2000-query budget (n=100) |
|---|---|---|
| LEAP success rate | 95.0% | 24.0% |
| PSO (WordNet) success rate | 95.0% | 23.0% |
| LEAP avg. queries | 44,071.8 | 1,540.8 |
| PSO (WordNet) avg. queries | 46,753.4 | 1,560.1 |
| LEAP avg. % words changed | 9.28% | 4.51% |
| PSO (WordNet) avg. % changed | 9.87% | 5.04% |
| LEAP wall time | 157.3s | 43.3s |
| PSO (WordNet) wall time | 353.6s | 46.5s |

LEAP matches or modestly beats this apples-to-apples PSO baseline on every metric, but the margin is small rather than dramatic, and narrows further under a realistic query budget — at 2000 queries, both methods' success rate collapses well below their unrestricted numbers, since neither search gets far enough in to fully exploit its strategy. Treat as directional evidence (single model/dataset/seed/sample size), not a definitive benchmark of LEAP vs. PSO in general.

## Test plan

- [x] `tests/test_attack_recipes.py` passes (structural + functional comparison against `PSOZang2020`)
- [x] CLI smoke tests (`textattack attack --recipe leap`, `--recipe pso`) against `cnn-ag-news`: both succeed with no errors, no regression in `pso`'s behavior after the shared base-class changes
- [x] `tests/benchmark_leap_vs_pso.py` run at two query-budget settings (see results above)
- [x] LEAP2023's docstring RST table verified to parse cleanly via docutils (no malformed-table errors)


qiyanjun:
Added a follow-up commit addressing three more issues surfaced during review:

- **Divide-by-zero guard**: `fit_ave`/`fit_min` are computed once from the initial population and never updated, but individual particle scores drift over later iterations — added a `fit_ave > fit_min` guard around the omega interpolation so a later score dipping below the frozen `fit_min` can't produce a zero/negative denominator.
- **Elite aliasing in `_turn`** (shared base class, `particle_swarm_optimization.py`, also affects `PSOZang2020`): when no constraint-passing move was found, `_turn` returned `source_text` itself — often literally `local_elites[k]` or `global_elite` — instead of a copy. `population[k]` could then become the same object as an elite, and a later in-place mutation (`_greedy_perturb`/`_perturb`) would silently corrupt that elite's tracked state. Now returns `copy.copy(source_text)`, consistent with the `copy.copy` pattern already used elsewhere in this file for elite updates. This was more likely to bite LEAP specifically, since softmax-normalized `turn_prob` (see below) makes the "no words selected to turn" no-op path more common than in vanilla PSO.
- **`softmax` vs. the parent's `sigmoid` for `turn_prob`**: checked this against the authors' reference implementation and it's an intentional design choice (LEAP normalizes turn probability across the whole sentence rather than treating each word independently, targeting ~1 word turned per call, consistent with the recipe's tight `MaxModificationRate(0.16)`), not a porting bug — left the behavior as-is but added a comment so it isn't "fixed" into sigmoid by someone reading only the parent class's docstring.

Re-verified with a smoke test (`cnn-ag-news`, 2 examples, unrestricted query budget): LEAP still succeeds 2/2 (100%), and a quick vanilla-PSO run confirms the shared `_turn` fix doesn't break the existing `pso` recipe.

qiyanjun:
Pushed one more commit:

- **`omega` reset per iteration**: `omega` was declared once before the `max_iters` loop and only ever appended to. Since the velocity update reads `omega[k]` for `k in range(pop_size)`, every iteration after the first was silently reusing iteration-0's inertia weights instead of the freshly computed ones — disabling the adaptive-inertia mechanism that's LEAP's main distinction from vanilla PSO, for 19 of 20 iterations by default, and growing the list to 1200 mostly-dead entries. This is present in the authors' own reference implementation too, but chose to fix it here rather than reproduce it, since there's no reason to intentionally keep a non-adaptive "Adaptive PSO."
- **Docs**: fixed the `ParticleSwarmOptimizationLEAP` RST section underline (was 26 dashes for a 29-char title — docutils flags this as "Title underline too short").

Re-ran the smoke test (`cnn-ag-news`, 2 examples, unrestricted budget) after this change: still 2/2 successful attacks, no regression.

qiyanjun:
One more fix pushed: **capped the retry loop** in \`get_one_levy\`/\`v_init_levy\`. Both used unbounded \`while True\` rejection sampling against the heavy-tailed Levy(alpha=1.5) distribution with no cap. \`get_one_levy\` now bounds retries to 1000 attempts, falling back to a uniform draw within the target range on an exhausted streak; the \`v_init_levy\` fill loop now just calls \`get_one_levy\` \`pop_size\` times instead of duplicating the same unbounded loop inline.

Re-verified with the same smoke test (\`cnn-ag-news\`, 2 examples, unrestricted budget): still 2/2 successful attacks, no regression.

That closes out every issue raised in review so far — the mutation-logic bugs, divide-by-zero guard, elite-aliasing fix, omega reset, docs underline, and this retry cap.

qiyanjun:
Pushed tests, docs, and a changelog:

- **Tests** (\`tests/test_attack_recipes.py\`): a structural test confirming \`LEAP2023\`'s search method is a specialization of the same \`ParticleSwarmOptimization\` family \`PSOZang2020\` uses (shared pop_size/max_iters defaults, goal function), while differing in transformation (WordNet vs. HowNet) and constraints (\`MaxModificationRate\` vs. \`RepeatModification\`) as intended; plus a functional test running both recipes end-to-end on the same sample as a regression guard.
- **Docs/comments**: cross-linked \`LEAP2023\` and \`PSOZang2020\` in both recipes' docstrings, the search method's module/class docstrings, \`docs/3recipes/attack_recipes.rst\`, and — this one was missing entirely — added \`leap\` to the README's attack-recipe comparison table.
- **CHANGELOG.md**: none existed in the repo, so started one with an \`[Unreleased]\` section documenting the LEAP addition and this PR's fixes.

All new/existing tests pass locally.

qiyanjun:
Pushed a follow-up addressing the latest review pass's findings:

**Correctness — the earlier elite-aliasing fix was incomplete.** I'd only fixed the case where \`_turn\` returns \`source_text\` by reference on a failed constraint check. The actual root cause was upstream: \`global_elite = max(population, ...)\` and \`local_elites = copy.copy(population)\` never copied the \`PopulationMember\` objects themselves — only the list/reference. Whenever a particle's random P1/P2 draws both failed (so \`_turn\` was never even called that iteration, a real ~16–25% chance per particle per iteration), \`population[k]\` stayed the literal same object as its local/global elite from initialization, and the mutation step's in-place update then silently corrupted that elite. Now every population member and both elites are copied individually at initialization — in both the shared base class (also used by \`pso\`) and LEAP.

**De-duplication.** While fixing the above, factored \`ParticleSwarmOptimization.perform_search\`'s per-iteration deltas into overridable hooks (\`_initialize_velocities\`, \`_pre_iteration_setup\`, \`_compute_omega\`, \`_compute_turn_prob\`, \`_compute_change_ratio\`), matching the hook pattern \`GeneticAlgorithm\`/\`AlzantotGeneticAlgorithm\` already use elsewhere in this codebase. \`ParticleSwarmOptimizationLEAP\` now overrides only these hooks plus \`_perturb\` (folding in what used to be the separately-named \`_greedy_perturb\`, which is exactly how it ended up dead code originally), instead of duplicating a ~155-line copy of \`perform_search\` that could silently drift from the base class.

**Efficiency/hygiene.** Replaced the hand-rolled \`softmax\` with \`scipy.special.softmax\`; cached the alpha-invariant constants in the Levy-flight sampler (\`alpha\` is always \`1.5\` here) with \`functools.lru_cache\`; renamed \`levy()\`'s \`gamma\` parameter to \`scale\` since it shadowed the \`scipy.special.gamma\` import. Deliberately left the hand-rolled Levy-stable sampling algorithm itself alone rather than swapping to \`scipy.stats.levy_stable\` — that would change the search's statistical behavior (different sampling algorithm, different tail behavior) rather than just its implementation, same tradeoff as the earlier softmax-vs-sigmoid decision.

Verified with \`tests/test_attack_recipes.py\` (both recipes, still passing) and CLI smoke tests against \`cnn-ag-news\` for both \`leap\` and \`pso\`: LEAP still 2/2 successful attacks, \`pso\`'s behavior on the same examples is unchanged from pre-refactor baseline (no regression).

qiyanjun:
Pushed the LEAP-vs-PSO benchmark results into the codebase itself, not just this PR thread:

- **\`tests/benchmark_leap_vs_pso.py\`**: a reproducible (but manual — not CI-run, since a meaningful sample takes minutes) script comparing \`LEAP2023\` against a WordNet-transformation variant of vanilla \`ParticleSwarmOptimization\`. \`PSOZang2020\` normally uses HowNet, so comparing it to LEAP directly conflates "which candidate-word pool is better" with "which search algorithm is better" — this script holds transformation/constraints/hyperparameters fixed and varies only the search method, so the comparison actually isolates LEAP's algorithmic contribution.
- **\`LEAP2023\`'s docstring**: added a "Benchmark" section with the results table (\`cnn-ag-news\`, AG News test set, seed 765): unrestricted budget → both ~95% success (n=20), LEAP using ~6% fewer queries and ~2.25x faster wall time; 2000-query budget → both collapse to ~23-24% success (n=100), LEAP still modestly ahead on every metric but by a much smaller margin. Explicitly caveated as directional evidence (single model/dataset/seed/sample size), not a definitive benchmark.
- Cross-referenced from \`tests/test_attack_recipes.py\`'s module docstring and \`CHANGELOG.md\`.

Verified the docstring's RST table parses cleanly via docutils (no malformed-table errors) before pushing, and re-ran the existing tests — still passing.

qiyanjun:
Merged in \`master\` to resolve the conflict flagged earlier — \`master\` had drifted significantly since this branch was cut (\`RemoteModelWrapper\`, a new \`DifferentialEvolution\` search method, a docs restructure, the \`bad-characters\` recipe, and more).

Conflicts and how they were resolved:
- **\`CHANGELOG.md\`**: both sides added \`[Unreleased]\` entries — kept master's intro line, merged both \`### Added\` sections, kept this PR's \`### Changed\`/\`### Fixed\` sections.
- **\`docs/api/search_methods.rst\`, \`textattack/attack_args.py\`, \`textattack/search_methods/__init__.py\`**: both sides added independent entries (\`ParticleSwarmOptimizationLEAP\` vs. \`DifferentialEvolution\`; \`"leap"\` vs. \`"bad-characters"\` recipe keys) — kept both.
- **\`textattack/search_methods/particle_swarm_optimization.py\`**: auto-merged cleanly with no manual resolution needed — master's only change there was formatting already identical to what this branch had applied, so this PR's hook-based refactor applied on top with zero real conflict (verified by diffing the merged result against \`origin/master\`).

Also picked up a new required dependency from master (\`Levenshtein\`, for a new \`MaximizeLevenshtein\` goal function) and confirmed everything still imports and tests pass with it installed.

Re-verified post-merge: \`tests/test_attack_recipes.py\` passes, \`textattack list search-methods\` shows both \`pso\` and the new \`differential-evolution\`, and a \`textattack attack --recipe leap\` smoke run behaves consistently with pre-merge runs on the same example.

PR now shows as \`MERGEABLE\` (previously \`CONFLICTING\`). It's still \`BLOCKED\` on required status checks, but per the earlier check — 0 check-suites have run at all, which looks like the fork-PR "requires approval to run workflows" gate rather than a failure. A maintainer will need to approve running the workflows from the PR's checks tab.

qiyanjun:
## Is LEAP worth including, based on the benchmark?

Yes — but with a realistic framing rather than the paper's dramatic pitch.

**Why it's worth it:**
- **The gains are real, just modest.** Across both benchmark conditions (see the table above / `LEAP2023`'s docstring), LEAP matched or beat vanilla PSO on every metric — success rate (tied at 95%/95% unrestricted, 24%/23% budget-capped), query efficiency (~5-6% fewer queries), perturbation size (~6-10% fewer words changed), and wall time (2.25x faster unrestricted). That's a consistent, if unspectacular, edge — not noise in one direction and reversal in another.
- **Low integration/maintenance cost.** After the hook-based refactor, `ParticleSwarmOptimizationLEAP` is ~200 lines overriding 5 well-defined hooks on the existing `ParticleSwarmOptimization`, not a parallel ~450-line implementation to maintain separately. It reuses existing goal functions, constraints, and transformations.
- **It fits TextAttack's mission** — a library of reproducible attacks from the literature. LEAP is a natural, citable extension of the PSO family already represented by `PSOZang2020`, and now has test coverage `PSOZang2020` itself never had.
- **It's now correctly implemented**, which wasn't true of the original submission (#794) — the mutation-logic, elite-aliasing, omega-staleness, and divide-by-zero bugs found and fixed during this review would have made the merged recipe silently behave like vanilla PSO, or occasionally corrupt its own search state. That gap between the paper's claims and what the code actually did is now closed and verified against the authors' own reference implementation.

**Where I'd push back on overselling it:**
- Under a realistic query budget (2000), LEAP and vanilla PSO are close to statistically indistinguishable (24% vs 23% on n=100) — the "efficient" framing in the paper's title is the part I'd be most skeptical of in practice.
- Both methods are still extremely query-hungry in absolute terms (tens of thousands of queries for high success rate) — a shared limitation of the whole PSO family here, not something LEAP fixes.
- Only validated on one model/dataset (`cnn-ag-news`/AG News) — I wouldn't claim the effect generalizes without testing elsewhere.

**Net**: include it, but the benchmark docstring's honest "directional, not definitive" framing is the right way to represent it — not "LEAP beats PSO," more "LEAP is a slightly-better-tuned PSO variant, now available and correctly implemented."

PR #831, opened 2026-08-13
Fix six more issues from the unresolved-bugs backlog

### LINKED ISSUE #722: Compatibility issue w/ Transformers model
When running the code example shown in the Attack class' docstring, the following error occurs:
`textattack: Unknown if model of class <class 'transformers.models.bert.modeling_bert.BertForSequenceClassification'> compatible with goal function <class 'textattack.goal_functions.classification.untargeted_classification.UntargetedClassification'>.`
**To Reproduce**
run the following code: 
```
import textattack
from textattack.constraints.pre_transformation import RepeatModification, StopwordModification
from textattack.constraints.semantics import WordEmbeddingDistance
from textattack.transformations import WordSwapEmbedding 
from textattack.search_methods import GreedyWordSwapWIR
from textattack import Attack
import transformers

# Load model, tokenizer, and model_wrapper
model = transformers.AutoModelForSequenceClassification.from_pretrained("textattack/bert-base-uncased-imdb")
tokenizer = transformers.AutoTokenizer.from_pretrained("textattack/bert-base-uncased-imdb")
model_wrapper = textattack.models.wrappers.HuggingFaceModelWrapper(model, tokenizer)

# Construct our four components for `Attack`
goal_function = textattack.goal_functions.UntargetedClassification(model_wrapper)
constraints = [
    RepeatModification(),
    StopwordModification(),
    WordEmbeddingDistance(min_cos_sim=0.9)
]
transformation = WordSwapEmbedding(max_candidates=50)
search_method = GreedyWordSwapWIR(wir_method="delete")

# Construct the actual attack
attack = Attack(goal_function, constraints, transformation, search_method)                                                   

input_text = "I really enjoyed the new movie that came out last month."
label = 1 #Positive
attack_result = attack.attack(input_text, label)
```
This code is sourced from the documentation of the Attack class.

See error.

**Expected behavior**
This code should run the attack and store the result in the attack_result variable.

**System Information (please complete the following information):**
 - OS: Linux
 - Textattack version: 0.3.8

**Additional context**
In shared/validators.py, the following code defines the globs which are used to validate compatibility between models and goal_functions:

```
MODELS_BY_GOAL_FUNCTIONS = {
    (TargetedClassification, UntargetedClassification, InputReduction): [
        r"^textattack.models.helpers.lstm_for_classification.*",
        r"^textattack.models.helpers.word_cnn_for_classification.*",
        r"^transformers.modeling_\w*\.\w*ForSequenceClassification$",
    ],
    (NonOverlappingOutput, MinimizeBleu,): [
        r"^textattack.models.helpers.t5_for_text_to_text.*",
    ],
}
```
However, the following glob is incorrect:
` r"^transformers.modeling_\w*\.\w*ForSequenceClassification$"`

The correct module path for models in Transformers is as follows:
` r"transformers.models.\w*.modeling_\w*.\w*ForSequenceClassification"`

**Suggested fix:**

Add the correct module path to the list of globs like so:
```
MODELS_BY_GOAL_FUNCTIONS = {
    (TargetedClassification, UntargetedClassification, InputReduction): [
        r"^textattack.models.helpers.lstm_for_classification.*",
        r"^textattack.models.helpers.word_cnn_for_classification.*",
        r"^transformers.modeling_\w*\.\w*ForSequenceClassification$",
        r"transformers.models.\w*.modeling_\w*.\w*ForSequenceClassification"
    ],
    (NonOverlappingOutput, MinimizeBleu,): [
        r"^textattack.models.helpers.t5_for_text_to_text.*",
    ],
}
```
This fixed the issue for me, let me know if I missed anything! 

I have a branch with this fix already, so let me know if I can help/make a PR.

### LINKED ISSUE #723: Bug of utils/string.py/words_from_text()
Hi,

In T=this function words_from_text() in utils/string.py (used in AttackText for generating words list of a given sentence), 

It seems that the function has difficulty handling quoted words.

For example, if an input is " AAA BBB 'CCC' ", the function words_from_text() returns AAA BBB and CCC', the left quote is deleted but the right quote is still reserved.

 I would appreciate it if you could kindly take a look at this and see if there is a way to resolve the issue.

### LINKED ISSUE #728: Issue with using AttackArgs class outside of the command line (in my Python script)
Hello,

I am experiencing an issue when trying to use the AttackArgs class from the TextAttack library outside of the command line. I'm attempting to use the class programmatically in my Python script, but the arguments I provide are not correctly assigned.

Here's a code snippet demonstrating my problem:

```
from textattack.attack_args import AttackArgs

attack_args = AttackArgs(
num_examples=100, 
num_successful_examples=2, 
log_to_csv="text_attack_results.csv"
)```

print(attack_args)


When I run the code, the **num_examples** values are not updated and remain at their default values. However, it seems that other arguments like **num_successful_examples** or **log_to_csv** are being correctly assigned.

I suspect this might be related to the class implementation, specifically the _add_parser_args method, which seems to be designed for command-line usage. Is there a way to use the AttackArgs class programmatically in a Python script without relying on the command line interface? Any guidance or suggestions would be appreciated.

Thank you for your assistance.

System Information:
 - OS: MacOS Ventura 13.1
 - Textattack version: 0.3.8

### LINKED ISSUE #754: a2t bug
**Describe the bug**
image

**To Reproduce**
1. Run following command `textattack ...`
2. Run following code ...

**Expected behavior**
It ask us to input truncate_words_to
**Screenshots or Traceback**
If applicable, add screenshots to help explain your problem. Also, copy and paste tracebacks produced by the bug. 
image

**System Information (please complete the following information):**
 - OS: [e.g. MacOS, Linux, Windows]
 - Library versions (e.g. `torch==1.7.0, transformers==3.3.0`)
 - Textattack version

**Additional context**

### LINKED ISSUE #771: Issues attacking se2seq model (seq2sick)
Seems like raw outputs of the model are being passed to `NonOverlappingOutput` implemented methods `_is_goal_complete` which subsequently calls `word_difference_score` with those tensors leading to` split()` being called on tensors giving `TypeError: split() missing 1 required positional argument: 'split_size'`. Let me know if there is something wrong while constructing the `model_wrapper`.

### LINKED ISSUE #787: the return of AttackedText.word_diff_ratio is 1 all the time

### LINKED ISSUE #798: Attack on HF models split into multiple GPUs
Does textattack allow running the attack calculation on a single model doing inference across multiple GPUs?

For example, one can use the argument `device_map="auto"`, which distributes a large model to multiple GPUs, on `HuggingFaceModelWrapper`. However, it seems that if you split a single instance of a large model onto multiple GPUs, then when doing the attack, such as using `attacker.attack_dataset()`, then there will be a RuntimeError similar to the following (if two GPUs are present)
`RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1!`

### LINKED ISSUE #800: `BackTranslationAugmenter` produces an incorrect number of transformations
**Describe the bug**
Calling the `augment` function of a `BackTranslationAugmenter` produces a list with a length that appears to be non-deterministic, sometimes not matching the value of the `transformations_per_example` initialization parameter.

**To Reproduce**
Run the following code:
```
import textattack
augmenter = textattack.augmentation.recipes.BackTranslationAugmenter(transformations_per_example=5)
s = "You are in the kitchen."
ls = augmenter.augment(s)
print(len(ls))
```

**Expected behavior**
Expected output: 5
Actual output: 3

**System Information (please complete the following information):**
 - OS: MacOS Sonoma 14.6.1
 - Library versions: torch==2.3.1, transformers==4.42.4
 - Textattack version: textattack==0.3.10
 - Full `pip freeze` output: https://pastebin.com/VPt7JdRY

### PR DESCRIPTION
**Pull request by qiyanjun**
## Summary

Second batch from a triage pass over the 24 open bugs in the issue backlog (first batch: #830). Each fix below is verified against the actual current code, not just the original report — three issues in this batch (#712/#715/#717/#730/#743/#753, listed below) turned out to already be fixed by unrelated prior commits once checked against `master`, so no code change was needed for those; comments were left on those issues instead.

- **`GreedyWordSwapWIR`**: actually implement `truncate_words_to`. `a2t`'s recipe (PR #747) has been passing this kwarg since 2023, but the constructor never declared it, so any use of `A2TYoo2021`/a2t-mlm crashes on init with `TypeError: unexpected keyword argument`. Fixes #754.
- **`validators.py`**: the model-compatibility regex only matched the pre-4.x `transformers.modeling_<model>` module layout. `transformers` reorganized this to `transformers.models.<model>.modeling_<model>` years ago, so every current install trips a spurious "unknown model" warning. Fixes #722 (also the source of the warning noise reported in #742, though that issue's "always fails" symptom needs a repro to diagnose further — commented there).
- **`AttackArgs`**: warn instead of silently dropping `num_examples` when it's explicitly set alongside `num_successful_examples` (the latter overrides the former by design, per the docstring, but users had no way to know why their value vanished). Fixes #728.
- **`AttackedText.words_diff_ratio`**: `self.words != x.words` compares two Python lists and returns a single bool, not an elementwise mask — so this always returned exactly 0 or 1 regardless of how many words actually differ. Fixed by comparing as numpy arrays. Fixes #787.
- **`Augmenter.augment`**: bounded-retry the outer sampling loop. It accumulates results in a `set()` but only ever attempted exactly `transformations_per_example` outer iterations; a transformation with limited output diversity for a given input (e.g. `BackTranslationAugmenter`'s randomly-chained back-translation converging to the same phrasing for short sentences) silently returned fewer unique results than requested. Fixes #800.
- **`words_from_text`**: strip allowed marks (quotes/hyphens/etc.) from both ends of a word, not just the leading end — a quoted word like `"'CCC'"` kept its trailing quote. Fixes #723.
- **`HuggingFaceModelWrapper`**: route encoder-decoder generation models (e.g. a raw `BartForConditionalGeneration` loaded directly via `transformers`, exactly as shown in TextAttack's own `Attacker` docstring example) through `.generate()` + decode instead of a plain forward pass. A plain forward pass only returns logits, which crashes text-to-text goal functions (`NonOverlappingOutput`, `MinimizeBleu`) that expect a string (`TypeError: split() missing 1 required positional argument`, since `word_difference_score` calls `.split()` on what it assumes is a string). TextAttack's own `T5ForTextToText` helper is unaffected (verified — it has no `.config` attribute, so it's routed around, not through, the new branch). Fixes #771.
- **`Attack.cuda_`**: skip re-placing a model that already has an `hf_device_map` attribute (set by `transformers`/`accelerate` when loaded with `device_map=...` across multiple GPUs). Forcing `.to(device)` on an already-sharded model conflicts with its placement and throws `RuntimeError: Expected all tensors to be on the same device`. Fixes #798.

Already fixed on `master` by unrelated prior commits — verified, no change needed, issues commented/closed accordingly: #712 (QWERTY IndexError, fixed Nov 2023), #715 and part of #713 (flair `force_token_predictions`, fixed alongside the #727 POS-tag work), #717 (QWERTY crash on symbol-only input, same 2023 fix as #712), #730 (`--high_yield` CLI flag, already registered), #743 (PSO `KeyError: 'pos'`, same root cause as #727), #753 (PSO `_turn` signature, already matches call sites post the LEAP refactor in #829).

## Test plan
- [x] `pytest tests/test_attacked_text.py tests/test_augment_api.py` (non-network-dependent subset): all pass, including two new regression tests (`test_quoted_word` for #723, plus the existing suite covering `Augmenter`/`AttackedText` paths touched here)
- [x] `make lint` clean (black/isort/flake8, matching the pinned CI versions)
- [x] `HuggingFaceModelWrapper` fix manually verified end-to-end against a real tiny seq2seq model (`hf-internal-testing/tiny-random-bart`: confirms `.generate()`+decode now returns strings) and a real classification model (confirms unaffected, still returns logits) and a mock mimicking `T5ForTextToText`'s shape (confirms it's routed around the new branch, not broken by it)
- [x] `Attack.cuda_` fix manually verified with a mocked `torch.nn.Module.to`: a module with `hf_device_map` set is skipped, a normal module is still moved as before
- [ ] `AttackArgs`/`words_diff_ratio`/`words_from_text`/`GreedyWordSwapWIR` fixes are small and locally reasoned/tested via direct interpreter calls (shown in the linked issue comments), but don't yet have dedicated unit tests beyond the one added for #723 — could use follow-up test coverage

### DIFF
--- a/tests/test_attack_device_placement.py
+++ b/tests/test_attack_device_placement.py
@@ -1,0 +1,108 @@
+def _build_attack(model, tokenizer, transformation):
+    from textattack import Attack
+    from textattack.constraints.pre_transformation import (
+        RepeatModification,
+        StopwordModification,
+    )
+    from textattack.goal_functions import UntargetedClassification
+    from textattack.models.wrappers import HuggingFaceModelWrapper
+    from textattack.search_methods import GreedyWordSwapWIR
+
+    wrapper = HuggingFaceModelWrapper(model, tokenizer)
+    goal_function = UntargetedClassification(wrapper)
+    return Attack(
+        goal_function,
+        [RepeatModification(), StopwordModification()],
+        transformation,
+        GreedyWordSwapWIR(),
+    )
+
+
+def _model_and_tokenizer():
+    import transformers
+
+    model = transformers.AutoModelForSequenceClassification.from_pretrained(
+        "hf-internal-testing/tiny-random-bert"
+    )
+    tokenizer = transformers.AutoTokenizer.from_pretrained(
+        "hf-internal-testing/tiny-random-bert"
+    )
+    return model, tokenizer
+
+
+def test_cuda_skips_hf_device_map_model_but_moves_unrelated_module():
+    # Regression test: the `hf_device_map` skip in `Attack.cuda_`/`to_cuda`
+    # used to apply to any `torch.nn.Module` with a truthy `hf_device_map`
+    # attribute, not just `transformers.PreTrainedModel` instances. Since
+    # this visitor also traverses non-HuggingFace modules reachable from a
+    # Constraint/GoalFunction/Transformation, a coincidental attribute name
+    # collision would silently skip moving that module to the configured
+    # device. Confirm the HF model is still (correctly) skipped, while an
+    # unrelated module with the same attribute name is not.
+    from unittest.mock import patch
+
+    import torch
+
+    from textattack.transformations import WordSwapRandomCharacterDeletion
+
+    model, tokenizer = _model_and_tokenizer()
+    model.hf_device_map = {"": "cpu"}
+
+    transformation = WordSwapRandomCharacterDeletion()
+    transformation.some_unrelated_module = torch.nn.Linear(2, 2)
+    transformation.some_unrelated_module.hf_device_map = {"": "cpu"}
+
+    attack = _build_attack(model, tokenizer, transformation)
+
+    with patch.object(model, "to", wraps=model.to) as model_to_spy, patch.object(
+        transformation.some_unrelated_module,
+        "to",
+        wraps=transformation.some_unrelated_module.to,
+    ) as marker_to_spy:
+        attack.cuda_()
+
+    assert model_to_spy.called is False
+    assert marker_to_spy.called is True
+
+
+def test_cpu_skips_hf_device_map_model_but_moves_unrelated_module():
+    # Same guard as cuda_/to_cuda, added separately to cpu_/to_cpu.
+    from unittest.mock import patch
+
+    import torch
+
+    from textattack.transformations import WordSwapRandomCharacterDeletion
+
+    model, tokenizer = _model_and_tokenizer()
+    model.hf_device_map = {"": "cpu"}
+
+    transformation = WordSwapRandomCharacterDeletion()
+    transformation.some_unrelated_module = torch.nn.Linear(2, 2)
+    transformation.some_unrelated_module.hf_device_map = {"": "cpu"}
+
+    attack = _build_attack(model, tokenizer, transformation)
+
+    with patch.object(model, "cpu", wraps=model.cpu) as model_cpu_spy, patch.object(
+        transformation.some_unrelated_module,
+        "cpu",
+        wraps=transformation.some_unrelated_module.cpu,
+    ) as marker_cpu_spy:
+        attack.cpu_()
+
+    assert model_cpu_spy.called is False
+    assert marker_cpu_spy.called is True
+
+
+def test_cuda_moves_model_without_device_map():
+    from unittest.mock import patch
+
+    from textattack.transformations import WordSwapRandomCharacterDeletion
+
+    model, tokenizer = _model_and_tokenizer()
+    transformation = WordSwapRandomCharacterDeletion()
+    attack = _build_attack(model, tokenizer, transformation)
+
+    with patch.object(model, "to", wraps=model.to) as model_to_spy:
+        attack.cuda_()
+
+    assert model_to_spy.called is True
--- a/tests/test_attacked_text.py
+++ b/tests/test_attacked_text.py
@@ -81,7 +81,7 @@ class TestAttackedText:
         assert attacked_text.text_of_first_n_words(1) == "A"
         assert attacked_text.text_of_first_n_words(3) == "A person walks"
         # n beyond the text's word count clamps to the full text.
-        assert attacked_text.text_of_first_n_words(10**5) == attacked_text.text[:-1]
+        assert attacked_text.text_of_first_n_words(10 ** 5) == attacked_text.text[:-1]
 
     def test_big_window_around_index(self, attacked_text):
         assert (
--- a/tests/test_attacked_text.py
+++ b/tests/test_attacked_text.py
@@ -76,6 +76,13 @@ class TestAttackedText:
             == "A person walks up stairs into a room and sees beer poured from a keg and people talking"
         )
 
+    def test_text_of_first_n_words(self, attacked_text):
+        assert attacked_text.text_of_first_n_words(0) == ""
+        assert attacked_text.text_of_first_n_words(1) == "A"
+        assert attacked_text.text_of_first_n_words(3) == "A person walks"
+        # n beyond the text's word count clamps to the full text.
+        assert attacked_text.text_of_first_n_words(10**5) == attacked_text.text[:-1]
+
     def test_big_window_around_index(self, attacked_text):
         assert (
             attacked_text.text_window_around_index(0, 10 ** 5) + "."
--- a/tests/test_attacked_text.py
+++ b/tests/test_attacked_text.py
@@ -43,6 +43,14 @@ def hyphenated_text():
     return textattack.shared.AttackedText(raw_hyphenated_text)
 
 
+raw_quoted_text = "AAA BBB 'CCC'"
+
+
+@pytest.fixture
+def quoted_text():
+    return textattack.shared.AttackedText(raw_quoted_text)
+
+
 @pytest.fixture
 def attacked_text_pair():
     return textattack.shared.AttackedText(raw_text_pair)
@@ -191,6 +199,12 @@ class TestAttackedText:
             new_text.text
             == "person walks a very long way up stairs into a room and sees beer poured and people on the couch."
         )
+
+    def test_quoted_word(self, quoted_text):
+        # Regression test for https://github.com/QData/TextAttack/issues/723:
+        # a word wrapped in quote marks should have both quotes stripped,
+        # not just the leading one.
+        assert quoted_text.words == ["AAA", "BBB", "CCC"]
 
     def test_hyphen_apostrophe_words(self, hyphenated_text):
         assert hyphenated_text.words == [
--- a/tests/test_augment_api.py
+++ b/tests/test_augment_api.py
@@ -93,6 +93,61 @@ def test_deletion_augmenter():
     assert augmented_s in augmented_text_list
 
 
+def test_high_yield_scales_with_transformations_per_example():
+    # Regression test: the retry-bound fix for issue #800 (stop a
+    # low-diversity transformation from silently returning fewer than
+    # `transformations_per_example` unique augmentations) had an
+    # intermediate version whose outer loop exited as soon as the target
+    # count was reached. In `high_yield=True` mode a single outer
+    # iteration can add many results to the set at once, so that made
+    # output plateau around the same size regardless of
+    # `transformations_per_example` instead of scaling with it (~4-13x
+    # fewer results, verified against pre-regression output for the same
+    # input/seed range).
+    from textattack.augmentation import Augmenter
+    from textattack.transformations.word_swaps import WordSwapWordNet
+
+    s = "A person walks up stairs into a room and sees beer poured from a keg and people talking."
+
+    def unique_count(n):
+        augmenter = Augmenter(
+            transformation=WordSwapWordNet(),
+            pct_words_to_swap=0.15,
+            transformations_per_example=n,
+            high_yield=True,
+        )
+        return len(set(augmenter.augment(s)))
+
+    small = unique_count(5)
+    large = unique_count(20)
+    # Loose bound (transformation output is stochastic): `large` should be
+    # meaningfully bigger than `small`, not roughly flat.
+    assert large > small * 2
+
+
+def test_augment_dedup_sample_from_set_no_crash():
+    # Regression test: the final downsampling step called
+    # `random.sample(all_transformed_texts, n)` where
+    # `all_transformed_texts` is a `set`. Python 3.11+ raises
+    # `TypeError: Population must be a sequence` for a set argument, since
+    # `random.sample` stopped accepting arbitrary sized iterables/sets.
+    # `fast_augment=True, high_yield=False` is what exercises this
+    # particular downsampling branch.
+    from textattack.augmentation import Augmenter
+    from textattack.transformations.word_swaps import WordSwapWordNet
+
+    augmenter = Augmenter(
+        transformation=WordSwapWordNet(),
+        pct_words_to_swap=0.2,
+        transformations_per_example=3,
+        high_yield=False,
+        fast_augment=True,
+    )
+    s = "A person walks up stairs into a room and sees beer poured from a keg and people talking."
+    augmented_text_list = augmenter.augment(s)
+    assert len(augmented_text_list) <= 3
+
+
 def test_high_yield_fast_augment():
     from textattack.augmentation import WordNetAugmenter
 
--- a/tests/test_huggingface_model_wrapper.py
+++ b/tests/test_huggingface_model_wrapper.py
@@ -1,0 +1,161 @@
+def test_generate_default_max_length_omitted():
+    # Regression test: the `.generate()` branch added for raw encoder-
+    # decoder generation models (#771) used to pass no length control at
+    # all, always falling back to whatever transformers/the model's own
+    # generation_config decided. `max_length=None` (the default) should
+    # leave `max_length` out of the `.generate()` call entirely, so a
+    # checkpoint with its own sensible generation_config isn't overridden.
+    import transformers
+
+    from textattack.models.wrappers import HuggingFaceModelWrapper
+
+    model = transformers.AutoModelForSeq2SeqLM.from_pretrained(
+        "hf-internal-testing/tiny-random-t5"
+    )
+    tokenizer = transformers.AutoTokenizer.from_pretrained(
+        "hf-internal-testing/tiny-random-t5"
+    )
+    wrapper = HuggingFaceModelWrapper(model, tokenizer)
+
+    captured = {}
+    original_generate = model.generate
+
+    def spy_generate(*args, **kwargs):
+        captured.update(kwargs)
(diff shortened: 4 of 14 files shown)

### CODE REVIEW

**Comment by qiyanjun** [comment]
## Comprehensive review

Reviewed the full diff, traced callers for each hunk, and empirically verified the two most severe findings against master. Ordered most → least severe.

### 1. `Augmenter.augment` (`textattack/augmentation/augmenter.py:130`) — `high_yield` output severely under-counts, confirmed
The new outer loop replaces `for _ in range(transformations_per_example)` with `while len(all_transformed_texts) < transformations_per_example`. In `high_yield=True` mode, a single outer iteration can add *many* `ready_texts` to the set at once, so the new early-exit can stop after just 1-2 iterations — whereas the old code always ran the full `transformations_per_example` outer iterations regardless of how many had already accumulated.

Verified directly (`WordSwapWordNet`, `pct_words_to_swap=0.15`, `high_yield=True`, same 17-word sentence):
| `transformations_per_example` | master (pre-PR) | this branch |
|---|---|---|
| 5 | 417 | ~104 |
| 10 | 786 | ~98 |
| 20 | 1326 | ~100 |

Master scales roughly linearly with `n` as intended; this branch is flat around ~100 regardless of `n` — a 4-13x reduction, and yield no longer scales with `transformations_per_example` at all. This looks like an unintended side effect of the #800 retry-bound fix, not the fix itself.

### 2. `HuggingFaceModelWrapper.__call__` (`huggingface_model_wrapper.py:77`) — missing `max_length` on `.generate()`, confirmed
The new `.generate()` call passes only `input_ids`/`attention_mask`, no length control, so it falls back to HF's generation defaults. Compare `textattack/models/helpers/t5_for_text_to_text.py:63`, which explicitly passes `max_length=self.output_max_length` for exactly this reason. A raw `BartForConditionalGeneration`/`T5ForConditionalGeneration` wrapped for `MinimizeBleu`/`NonOverlappingOutput` on inputs whose expected output exceeds the default will get silently truncated output, corrupting the BLEU/overlap comparison against `ground_truth_output`.

### 3. `AttackArgs.__post_init__` (`attack_args.py:229`) — warning gate misses the default value, confirmed
`self.num_examples not in (None, 10)` treats an explicit `num_examples=10` identically to "unset" (10 is the dataclass default). A user who explicitly passes `num_examples=10` alongside `num_successful_examples` — a very plausible value to pick — gets no warning even though `num_examples` is still silently overridden to `None`, defeating the #728 fix for that case.

### 4. `HuggingFaceModelWrapper.__call__` — encoder-decoder branch checks architecture, not task head (lower confidence)
`getattr(model_config, "is_encoder_decoder", False) and hasattr(self.model, "generate")` doesn't distinguish a generation model from a seq2seq-backbone classification model (e.g. `BartForSequenceClassification`, whose config also sets `is_encoder_decoder=True`). On the currently pinned transformers (4.57.1) `hasattr(....), "generate")` is False for such models so this doesn't currently trigger, but `requirements.txt` pins only `transformers>=4.30.0` with no upper bound — on older versions where `GenerationMixin` wasn't yet gated by `can_generate()`, this could misroute a classification model into `.generate()` and crash or return garbage instead of logits.

### 5. `GreedyWordSwapWIR` (`greedy_word_swap_wir.py:48`) — `truncate_words_to` doesn't bound `wir_method="gradient"` cost (lower confidence)
The truncation slices `indices_to_order` *after* `get_indices_to_order` runs, but for `wir_method="gradient"` the expensive step (`victim_model.get_grad(...)` on the full untruncated text) still runs at full cost — only the cheap post-hoc index-scoring loop is shortened. `A2TYoo2021.build()` passes `truncate_words_to=max_len` specifically to bound cost for the gradient method against a model's context window, so the fix may not actually bound cost the way it's intended to for that caller.

### 6. `Attack.cuda_`/`to_cuda` (`attack.py:215`) — device-map skip layered too generically (style/design)
The `hf_device_map` skip lives in the generic, framework-agnostic `to_cuda` visitor rather than being scoped to HuggingFace models specifically. Any non-HF `nn.Module` reachable from a Constraint/GoalFunction/Transformation that happens to carry a truthy `hf_device_map` attribute would now also be skipped, since the check only tests attribute presence.

### 7. `Augmenter.augment` — `max_attempts = transformations_per_example * 3` (minor)
Unexplained magic multiplier hardcoded inline. A transformation with less output diversity than BackTranslation (the case the comment cites) can still exhaust 3x attempts without reaching the target, with no way for a caller to raise the retry budget.

---
Findings 1-3 are empirically/structurally confirmed; 4-5 are plausible but lower-confidence (version/caller-dependent); 6-7 are design nits, not correctness bugs.

**Comment by qiyanjun** [comment]
## Code review (`/code-review`)

Ran a review over the current diff and found 5 issues, all verified by direct reproduction or code inspection. All 5 have since been fixed on the branch.

### 1. `textattack/augmentation/augmenter.py` — `random.sample()` on a `set` crashes on Python 3.11+
`random.sample(all_transformed_texts, ...)` was called directly on `all_transformed_texts`, a `set`. Python 3.11+ removed `random.sample` support for sets, so this raised `TypeError: Population must be a sequence...`.

Repro (pre-fix):
```python
WordNetAugmenter(pct_words_to_swap=0.5, transformations_per_example=2, fast_augment=True, high_yield=False).augment("A cat sat on the mat and the dog ran fast")
# TypeError: Population must be a sequence. For dicts or sets, use sorted(d).
```
**Fix:** wrap in `list(...)` before sampling.

### 2. `textattack/attack.py` — `cpu_()` missing the `hf_device_map` guard `cuda_()` has
`to_cuda` (added to fix #798) skips `.to(device)` for `transformers.PreTrainedModel`s loaded with `device_map=...` (accelerate-sharded across multiple GPUs), since forcing them onto one device breaks that placement. `to_cpu` had no equivalent guard — it unconditionally called `.cpu()` on every `torch.nn.Module`, so moving an `Attack` to CPU (e.g. to free memory between attacks) still collapsed a sharded model's placement.

**Fix:** mirrored the same `hf_device_map` guard into `to_cpu`.

### 3. `textattack/augmentation/augmenter.py` — `high_yield` mode regressed to ~half its previous yield
The outer loop was changed from `for _ in range(transformations_per_example)` to `while len(all_transformed_texts) < transformations_per_example and num_attempts < max_attempts` (as part of the #800 fix for low-diversity transformations like BackTranslation). That change made the loop exit as soon as the target count was hit, instead of always completing `transformations_per_example` full passes — which broke `high_yield` mode's intended over-generation behavior (a single pass can add several results at once).

A/B comparison of old vs. new loop logic on identical seeds showed the new logic consistently returning 30–60% fewer unique augmentations (e.g. `'A cat sat on the mat'`, seed=1: old=54, new=17). `test_high_yield_fast_augment` only asserts `>=` between modes, not an absolute count, so this went untested.

**Fix:** changed the while-condition so the loop always runs at least `transformations_per_example` attempts (restoring `high_yield`'s over-generation) while still allowing extra bounded retries past that if the target hasn't been reached (preserving the #800 fix), capped by `max_attempts`.

### 4. `textattack/search_methods/greedy_word_swap_wir.py` — `truncate_words_to` assumed set iteration order is ascending
`indices_to_order` comes from a `set` intersection in `Transformation.__call__` (`indices_to_modify & constraint(...)`), then `list(...)`. CPython set iteration order isn't guaranteed to ascend by value for arbitrary integers — verified directly (e.g. `set([485, 2924, 2524, 2160, 4818, 4081, 26])` iterates as `[485, 2924, 2160, 4081, 4818, 26, 2524]`). `indices_to_order[:truncate_words_to]` therefore didn't reliably return "the first N word positions" as documented, which could leave `last_index = max(indices_to_order)` in the `wir_method="gradient"` branch spanning most of the original text — defeating the cost bound `a2t_yoo_2021.py` relies on this for.

**Fix:** sort before truncating: `sorted(indices_to_order)[: self.truncate_words_to]`.

### 5. `textattack/attack_args.py` — CLI default defeated the `num_examples`/`num_successful_examples` conflict warning
(From an earlier round of this same review pass, already fixed before this comment.) The `--num-examples` argparse default was `default_obj.num_examples`, i.e. the sentinel already resolved to the concrete `10`. Any CLI user who only passed `--num-successful-examples` (never touching `--num-examples`) got a spurious "conflicting" warning, since `AttackArgs.__post_init__` couldn't distinguish "explicitly passed 10" from "never touched." Fixed by pointing the argparse default at the raw unresolved sentinel instead.

---
All 5 fixes verified: `tests/test_augment_api.py` passes (`test_high_yield_fast_augment` included), plus targeted reproductions for each finding confirming the fix.

🤖 Posted via Claude Code

qiyanjun:
**Note for future reference: local `test_back_translation`/`test_back_transcription` native crash — pre-existing, unrelated to this PR, likely dev-machine-specific**

While adding test coverage for this PR's fixes, `tests/test_augment_api.py::test_back_translation` and `test_back_transcription` reliably segfaulted on my local machine (macOS, Python 3.13). Root-caused it — it's a genuine native-library ABI collision, not a TextAttack bug:

- `import flair` (triggered transitively by `import textattack`, since `textattack/shared/utils/strings.py` imports flair at module scope) followed by loading a `sentencepiece`-based tokenizer (`transformers.MarianTokenizer`, used by `BackTranslation`) → immediate crash: `libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument`.
- Reversing the order (load the tokenizer first, `import flair` after) doesn't help either — it deadlocks instead, hanging forever inside `[mutex.cc:452] RAW: Lock blocking` (abseil's own low-level mutex code).

Both `sentencepiece` and something in `flair`'s dependency chain independently bundle their own copy of Google's abseil library, and the two collide at the native/ABI level regardless of load order. This isn't fixable via a Python-side import reorder or lazy-import — it's a version-compatibility problem between compiled C++ extensions, specific to this combination of pinned `sentencepiece`/`flair`/`transformers` wheel versions on this OS/Python build.

Not fixed here — no in-repo code change addresses it. Flagging in case CI (or another contributor's machine) ever shows similar flakiness on these two tests: it's very likely this same class of issue, not a regression from this PR's changes. Confirmed via bisection: `sentencepiece` alone works, `transformers.MarianTokenizer` alone works, only the flair+sentencepiece combination breaks.

qiyanjun:
**Test coverage added for every commit on this PR**

Verified every fix directly — walked each commit's diff, and for the ones with any real risk of stale/incorrect behavior, cross-checked against pre-fix code to confirm the fix actually changes behavior (not just cosmetic).

Added 24 new regression tests across 4 new files + 2 extended existing files, covering every previously-untested fix in the PR:
- `test_search_methods.py` — sorted-indices truncation, gradient-cost bounding, paired-input safety
- `test_validators.py` — both compatibility regexes, no spurious warning
- `test_huggingface_model_wrapper.py` — `max_length` forwarding, `can_generate()` routing (including a scenario simulating the actual pre-`can_generate()` failure mode), `T5ForTextToText` path
- `test_attack_device_placement.py` — `cuda_`/`cpu_` `hf_device_map` scoping
- `test_augment_api.py` — `high_yield` scaling, `random.sample`-on-set crash
- `test_attacked_text.py` — `text_of_first_n_words` edge cases

Every new test was checked both ways: passes on current code, and — checked out against the pre-fix version of its target file — actually fails, confirming each one is a real regression guard rather than a test that would pass regardless.

Issue #824, opened 2026-05-11
Companion package proposal: textattack-detection plugin (content-layer defense recipe, out-of-tree)


eeee2345:
Hi @vlwk + @qiyanjun — I know TextAttack has been mostly in maintenance mode since v0.3 and that Jack moved to Cornell. So I'm not asking for any core changes; this proposal is **deliberately shaped as an out-of-tree companion package** that depends on TextAttack and adds a detection lane.

The proposal: a `textattack-detection-atr` PyPI package (MIT, no CLA) that exposes a `Detector` interface compatible with TextAttack's `Attacker` and a reference implementation backed by ATR rules. Users opt in via:

```python
pip install textattack-detection-atr
```

```python
from textattack_detection_atr import ATRDetectionRecipe
from textattack import Attacker

attack = TextFoolerJin2019.build(model_wrapper)
attacker = Attacker(attack, dataset, detector=ATRDetectionRecipe())
result = attacker.attack_dataset()

# Existing TextAttack metrics unchanged
print(result.attack_success_rate)

# New
print(result.bypass_rate)  # ASR among attacks the detector didn't catch
```

Zero changes to `QData/TextAttack`. You don't approve anything, you don't maintain anything, your CI doesn't take on test surface. The only ask is:

- **A line in the README** under a "Defense / detection extensions" section pointing at the companion package, so users can discover it.

That's it.

## Why this exists

A meaningful chunk of TextAttack's actual usage is in undergraduate / graduate NLP security curricula. Students run TextFooler / BERT-Attack / etc. against a victim model and conclude "the attack succeeded N% of the time." There's no built-in step that asks "and would a content-layer defense have caught it?" — which is the question every security course actually wants to answer.

A defense-evaluation companion module gives that question a one-line answer. It's the kind of extension that helps a foundational framework stay relevant in curricula even when active development has slowed.

## About the detector

[Agent-Threat-Rule/agent-threat-rules](https://github.com/Agent-Threat-Rule/agent-threat-rules) is MIT-licensed, 344 rules, 6-check pre-merge quality gate including 0-FP on a 1,941-sample benign + research-mention corpus. The text-classification rules in the pack cover prompt injection, jailbreak phrasings, and a chunk of the TextFooler / BERT-Attack canonical shapes — though honestly, the framework's strength against textual adversarial attacks is regex against well-known shapes, not novel paraphrases. We're upfront about this in the package README.

## What I'm asking

[...]

Issue #801, opened 2024-08-26
可以添加一个新的攻击方法吗？https://github.com/AdvAttack/BU-SPO


zl-comment:
因为想要使用https://github.com/AdvAttack/BU-SPO这个攻击方法去进行一个大模型的攻击，可以将这个写进来吗？如果要自己写的话，是否有很好的模块添加的路线？

qiyanjun:
please submit PR .. thanks

qiyanjun:
This is a full new attack recipe implementation (porting BU-SPO's algorithm from https://github.com/AdvAttack/BU-SPO into TextAttack's `Attack`/`Transformation`/`SearchMethod` structure), not something to implement blind without understanding the paper's method in depth — still open and welcome as a community PR, per the maintainer's original reply. See `CLAUDE.md`'s "Adding New Components" section for the recipe-contribution pattern (subclass `AttackRecipe`, implement `build(model_wrapper)`, register in `attack_recipes/__init__.py`).

Issue #799, opened 2024-08-23
Support Chinese attack？


SueeH:
#719 shows QData support chinese attack, the code merged to master branch？
if so， how to use chinese attack or augmentation？

qiyanjun:
Yes — Chinese support is merged and on \`master\` (\`textattack.attack_recipes.ChineseRecipe\`, using \`jieba\` for segmentation and Chinese-specific transformations like homophone/morphonym character swaps). To use it:

\`\`\`python
from textattack.attack_recipes import ChineseRecipe
attack = ChineseRecipe.build(model_wrapper)
\`\`\`

There's also a Chinese example notebook at \`docs/2notebook/Example_6_Chinese_Attack.ipynb\`. Note this covers *attacks*, not augmentation — there's no dedicated Chinese-language augmentation recipe currently (augmentation is English-focused), just the attack recipe.

qiyanjun:
Answered above — Chinese attack support is merged and usable via `ChineseRecipe`, with a usage example and pointer to the example notebook. Closing as answered.

PR #832, opened 2026-08-13
Fix training padding, add train --dataset-from-file, close doc gaps

### LINKED ISSUE #423: Could you add a list of the supported languages of textattack?
Hi, thanks for this amazing work.

I am wondering if you can specify the supported languages for each data augmenter? Or providing a list of the supported languages of `textattack` in general. That will be very helpful.

### LINKED ISSUE #625: Train model on custom dataset from file does not work
**Describe the bug**
Train model on custom dataset returns failure.

**To Reproduce**
1. Run following command `!textattack train --model distilbert-base-uncased --dataset-from-file <*.py> --model-num-labels 2 --model-max-length 64 --per-device-train-batch-size 128 --num-epochs 3`

**Expected behavior**
Expected to train the model from the custom dataset

**Screenshots or Traceback**
image

**System Information (please complete the following information):**
 - OS: Windows
 - Textattack version: latest

### LINKED ISSUE #673: Can you provide a simple attack API for single sentence attack test?
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Here is a simply idea for single text attack which is rewrited from [original batch attack function](https://github.com/yangheng95/TextAttack/blob/e650172d260db90a95ff1a056a9d5522b3f990b6/textattack/attacker.py#L129)
```python3
    def simple_attack(self, text, label):
        if torch.cuda.is_available():
            self.attack.cuda_()

        example, ground_truth_output = text, label
        try:
            example = textattack.shared.AttackedText(example)
            if self.dataset.label_names is not None:
                example.attack_attrs["label_names"] = self.dataset.label_names
            try:
                result = self.attack.attack(example, ground_truth_output)
            except Exception as e:
                raise e
            if (isinstance(result, SkippedAttackResult) and self.attack_args.attack_n) or (
                not isinstance(result, SuccessfulAttackResult)
                and self.attack_args.num_successful_examples
            ):
                return
            else:
                return result
        except KeyboardInterrupt as e:
            raise e
```
**Describe alternatives you've considered**
No

**Additional context**
I also suggest to support mirror the tfhub which will help people in China(unable to access tfhub.dev).
e.g., https://github.com/yangheng95/TextAttack/blob/e650172d260db90a95ff1a056a9d5522b3f990b6/textattack/constraints/semantics/sentence_encoders/universal_sentence_encoder/universal_sentence_encoder.py#L19
and https://github.com/yangheng95/TextAttack/blob/e650172d260db90a95ff1a056a9d5522b3f990b6/textattack/constraints/semantics/sentence_encoders/universal_sentence_encoder/multilingual_universal_sentence_encoder.py#L21 .
However, this is up to you and thanks very much!

### LINKED ISSUE #737: Improve training speed by allow padding to batch-wise max length instead of model supported max length
**Is your feature request related to a problem? Please describe.**
I'm running A2T augmented training with codes from QData/TextAttack-A2T. As mentioned in [QData/Textattack-A2T Issue#1](https://github.com/QData/TextAttack-A2T/issues/1), the training speed of the PLM model is extremely slow.

After some checking I notices in `textattack.Trainer.training_step()`, the input text is padded to the max length the model supported, instead of the max length in the batch. This causes much more computation overhead, and probably unnecessary.

I wrote an inherited Trainer class to override the `training_step` method which simply changed the `padding` parameter of tokenizer to `True` and the training speed is much faster.

**Describe the solution you'd like**
Improve the training speed of `transformers` models when using `textattack.Trainer`.
Previous test shows that on a 3090 card, when training SNLI task a clean epoch took 3 hours.

**Describe alternatives you've considered**
By allow the tokenizer to only pad the input text to the max length in the batch instead of the max length supported by the corresponding model. By doing so a clean epoch took 20 minutes on the same condition.

**Additional context**
Is there any specific necessary to pad the input text to the max length supported by model?

### LINKED ISSUE #772: Issue of  attacking seq2seq model
Hello, I've read issue #725 and the API doc, but only find one commandLine example using seq2sick. [link](https://textattack.readthedocs.io/en/master/3recipes/attack_recipes_cmd.html)
I need to attack BART in en-de translation task by using seq2sick and morpheus .
It would be great if you could provide the links for those examples in documentation, Thanks!

qiyanjun:
## Summary

Third batch from the ongoing issue-triage pass, covering enhancement/documentation issues where investigation turned up a real, boundedly-scoped fix (as opposed to a large new feature — those were left as comments, not implemented, see below).

- **`Trainer.training_step`/`evaluate_step`**: switch from `padding="max_length"` (pads every batch to the tokenizer's static max length) to `padding=True` (pads to the longest sequence in the batch). This was reported and independently validated by the reporter in their own `Trainer` subclass. Fixes #737.
- **`train` command**: add `--dataset-from-file` support. This was a real missing feature, not a latent bug — unlike `attack`/`eval` (which use the shared `DatasetArgs` with full `dataset_from_file` support), `train`'s own arg parser (`_CommandLineTrainingArgs`) never registered the flag at all. Now a Python module can define `train_dataset`/`eval_dataset` (each a `textattack.datasets.Dataset`) and be loaded via `--dataset-from-file path.py` (optionally `path.py^prefix` for `prefix_train_dataset`/`prefix_eval_dataset`). Fixes #625.
- **Docs**: 
  - Document `Attack.attack(text, label)` — the single-example API already exists (it's what `Attacker` calls per-example internally), it just wasn't in the Quick Tour, so users kept assuming they needed a full `Dataset`. Fixes #673.
  - Add a working Python example to `Seq2SickCheng2018BlackBox`'s docstring showing a raw `BartForConditionalGeneration` attack end-to-end, now that #771 (in #831) makes this path actually work. Fixes #772.
  - Add a "Multi-lingual attacks" section listing the three language-specific recipes (`FrenchRecipe`, `SpanishRecipe`, `ChineseRecipe`) that existed but were never documented. Fixes #423.

**Deliberately not implemented** (left as informative comments on the issues instead, since they're substantial new features/design decisions rather than bounded fixes): #636 (timeout mechanism — real feature, needs a design decision on new result types the maintainers themselves didn't converge on), #348 (save all transformed queries — new logger feature), #407 (paraphrase/sentence-level attacks — new attack category), #801 (BU-SPO recipe port — new recipe implementation), #444 (expose victim-model task metadata to transformations — API design question), #558 (WordDeletion + Augmentation `modified_indices` — verified current behavior is no longer *wrong*, just not fully tracking deletions; commented with what I verified rather than guessing at a redesign), #687 (adversarial defense integration — pointed at the relevant open proposal, #824), #731 (noted LEAP (2023) was already added via #829, so "no recipes since 2021" is now stale), #799/#790/#789/#775/#773/#732 (question/support issues — answered directly in comments where I had a confident answer, or asked for the specific info needed where I didn't).

## Test plan
- [x] `pytest tests/test_attacked_text.py tests/test_augment_api.py` (non-network-dependent subset): all pass
- [x] `make lint` clean (black/isort/flake8, pinned CI versions)
- [x] `--dataset-from-file` verified end-to-end against a fake module exposing `train_dataset`/`eval_dataset`
- [x] `textattack train --help` confirms `--dataset-from-file` is registered and the existing `--dataset` flow is untouched
- [ ] The `Trainer` padding change and doc-only changes don't have new automated tests (training is not covered by fast/offline tests in this environment) — logic verified by reading, not an end-to-end training run

### DIFF
--- a/docs/0_get_started/quick_api_tour.rst
+++ b/docs/0_get_started/quick_api_tour.rst
@@ -35,3 +35,17 @@ Next, let's build the attack that we want to use. TextAttack provides prebuilt a
 
 
 .. image:: ../_static/imgs/overview.png
+
+
+Attacking a single example
+------------------------------
+The :class:`~textattack.Attacker`/:class:`~textattack.datasets.Dataset` flow above is for running an
+attack over many examples with logging, checkpointing, and parallelism built in. If you just want to
+attack one piece of text, no dataset is needed — call ``.attack()`` directly on the ``Attack`` object
+returned by a recipe's ``build()`` (see https://github.com/QData/TextAttack/issues/673):
+
+.. code-block::
+
+    >>> attack = textattack.attack_recipes.TextFoolerJin2019.build(model_wrapper)
+    >>> result = attack.attack("I really enjoyed this movie.", 1)  # (text, ground_truth_label)
+    >>> print(result.__str__(color_method="ansi"))
--- a/docs/3recipes/attack_recipes.rst
+++ b/docs/3recipes/attack_recipes.rst
@@ -174,5 +174,32 @@ General
    :noindex:
 
 
+Multi-lingual attacks
+############################################
+
+TextAttack also includes recipes for non-English classification models, each a
+language-specific adaptation of a contextualized-perturbation attack (see
+https://github.com/QData/TextAttack/issues/423):
+
+20. FrenchRecipe -- Attack French Recipe (Contextualized Perturbation for French NLP Adversarial Attack)
+21. SpanishRecipe -- Attack Spanish Recipe (Contextualized Perturbation for Spanish NLP Adversarial Attack)
+22. ChineseRecipe -- Attack Chinese Recipe (Contextualized Perturbation for Chinese NLP Adversarial Attack)
+
+
+.. automodule:: textattack.attack_recipes.french_recipe
+   :members:
+   :noindex:
+
+
+.. automodule:: textattack.attack_recipes.spanish_recipe
+   :members:
+   :noindex:
+
+
+.. automodule:: textattack.attack_recipes.chinese_recipe
+   :members:
+   :noindex:
+
+
 
 
--- a/textattack/attack_recipes/seq2sick_cheng_2018_blackbox.py
+++ b/textattack/attack_recipes/seq2sick_cheng_2018_blackbox.py
@@ -28,6 +28,29 @@ class Seq2SickCheng2018BlackBox(AttackRecipe):
 
     This is a greedy re-implementation of the seq2sick attack method. It does
     not use gradient descent.
+
+    Works against any encoder-decoder generation model loaded via
+    :class:`~textattack.models.wrappers.HuggingFaceModelWrapper` (e.g. a raw
+    ``transformers.BartForConditionalGeneration`` or
+    ``transformers.T5ForConditionalGeneration``), not just TextAttack's own
+    ``T5ForTextToText`` helper. See
+    https://github.com/QData/TextAttack/issues/771 and
+    https://github.com/QData/TextAttack/issues/772.
+
+    Example, attacking a BART summarization model directly from
+    ``transformers``::
+
+        import transformers
+        from textattack.attack_recipes import Seq2SickCheng2018BlackBox
+        from textattack.models.wrappers import HuggingFaceModelWrapper
+
+        model = transformers.AutoModelForSeq2SeqLM.from_pretrained("facebook/bart-large-cnn")
+        tokenizer = transformers.AutoTokenizer.from_pretrained("facebook/bart-large-cnn")
+        model_wrapper = HuggingFaceModelWrapper(model, tokenizer)
+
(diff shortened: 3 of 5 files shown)


qiyanjun:
**Merge-order dependency on #831**

The docstring/example added here for `Seq2SickCheng2018BlackBox` (attacking a raw `BartForConditionalGeneration` directly via `HuggingFaceModelWrapper`) documents behavior that isn't actually implemented on this branch yet.

The fix that makes it work — routing encoder-decoder generation models through `.generate()` + `tokenizer.batch_decode()` instead of a plain forward pass — lives in a separate commit (`f21163be`, "HuggingFaceModelWrapper: route encoder-decoder generation models... through `.generate()`...") on PR #831, not in this PR. `textattack/models/wrappers/huggingface_model_wrapper.py` on this branch still does the old plain forward pass unconditionally.

If #832 merges before/without #831, the new docstring example will fail for anyone who copies it (a plain forward pass on a raw BART model returns logits, not decoded text, which breaks `NonOverlappingOutput`'s string comparison).

Suggest merging #831 first (or together with this one) so the documented example actually matches merged behavior.

